PR #84, opened 2026-01-16
fix(conductor): add negative examples to prevent false CRITICAL classifications

EivMeyer:
## Summary

Adds negative examples to junior-conductor and senior-conductor prompts to prevent false CRITICAL classifications for code refactoring tasks that mention deployment/config/infrastructure keywords.

**Problem:** Issue #880 (DI refactor) was classified as CRITICAL because the task mentioned "deployment configuration" - but this was TypeScript type refactoring, not actual infrastructure changes.

**Fix:** Added explicit examples showing that:
- "Refactor to use dependency injection pattern" → STANDARD
- "Extract service configuration into separate module" → STANDARD  
- "Implement factory pattern for executor creation" → STANDARD
- "Add deployment type discriminated union for type safety" → STANDARD

CRITICAL is ONLY for: actual auth logic, real billing code, production secrets, live infrastructure, real database migrations.

## Impact

- DI refactor tasks → STANDARD (2 validators) instead of CRITICAL (4 validators)
- Saves unnecessary validator overhead for code organization tasks
- No change to actual critical domain detection (auth/billing/secrets/infra)

## Test Plan

- [ ] Run classification on "Refactor to use DI pattern" → Should be STANDARD
- [ ] Run classification on "Update billing webhook" → Should still be CRITICAL


EivMeyer:
Superseded by new PR based on dev branch

PR #237, opened 2026-01-30
fix(daemon): exit process when cluster completes

EivMeyer:
## Problem

Zombie zeroshot processes remained after heroshot shipped PRs, blocking new agent spawns.

## Root Cause

When CLUSTER_COMPLETE triggers `orchestrator.stop()`, the cluster state changes to 'stopped' but no SIGTERM is sent. The cleanup handlers registered via `process.on('SIGTERM')` are never triggered.

## Fix

Added polling in `setupDaemonCleanup()` to detect when cluster state changes to 'stopped' or 'completed', then exit the process.

## Testing

The fix was triggered by a heroshot run where 3 zombie processes from already-shipped items (#1159, #1168) were blocking new spawns.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -574,21 +574,33 @@ See `src/logic-engine.js` for sandbox implementation.
 
 ### Context Building
 
-Agents build context from ledger messages before executing:
+Agents build context from ledger messages before executing. Use explicit selection semantics:
 
-```javascript
+```json
 {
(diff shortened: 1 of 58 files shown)


tomdps:

Closing this stale branch. It carries a large old release-sync history around a narrow daemon-exit fix, so it is not a clean integration path against current `dev`. Recreate the daemon-completion fix narrowly if it is still reproducible.

PR #468, opened 2026-04-07
Security: Path traversal/arbitrary file write risk in parallel copy worker

tuanaiseo:
## Problem

Destination paths are created using `path.join(destBase, relativePath)` without validating that `relativePath` stays within `destBase`. If `files` contains `../` segments or absolute-like traversal payloads, writes can escape the intended directory and overwrite arbitrary files.

**Severity**: `high`
**File**: `src/copy-worker.js`

## Solution

Normalize and validate each path before copy: resolve destination (`path.resolve(destBase, relativePath)`) and enforce it starts with `path.resolve(destBase) + path.sep`. Reject entries that fail this check.

## Changes

- `src/copy-worker.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/copy-worker.js
+++ b/src/copy-worker.js
@@ -14,10 +14,17 @@ const { files, sourceBase, destBase } = workerData;
 let copied = 0;
 let skipped = 0;
 const errors = [];
+const resolvedDestBase = path.resolve(destBase);
 
 for (const relativePath of files) {
   const srcPath = path.join(sourceBase, relativePath);
-  const destPath = path.join(destBase, relativePath);
+  const destPath = path.resolve(destBase, relativePath);
+
+  if (destPath !== resolvedDestBase && !destPath.startsWith(resolvedDestBase + path.sep)) {
+    skipped++;
+    errors.push({ file: relativePath, error: 'Invalid destination path' });
+    continue;
+  }
 
   try {
     // Ensure parent directory exists


tomdps:
Status review (2026-07-29):
- Current main lacks a general destination-containment invariant: `src/copy-worker.js:18-30` joins worker destinations and writes them without validating containment.
- Current supported tar/archive input provenance narrows demonstrated exploitability: no supported extractor feeds raw archive member paths into this copy pipeline; production names are local-directory `readdirSync(..., { withFileTypes: true })` entries built in `src/isolation-manager.js:969-1026`.
- PR #468 is not merge-ready: its worker-only check misses the small-copy sink (`src/isolation-manager.js:1047-1061`), directory creation (`src/isolation-manager.js:1028-1045`), and an explicit symlink policy (`src/isolation-manager.js:980-1002`); containment must cover all destination-producing paths.

tomdps:
Thanks for surfacing this boundary. I rechecked the current copy pipeline: the containment invariant is missing, but this worker-only lexical check would leave sequential directory creation, the small-copy path, and destination symlink escapes unprotected. Current input provenance also narrows the demonstrated arbitrary-write claim, so I don't want to merge a partial fix under the stronger security contract.

I've opened #939 to retain the report and define one source/destination and symlink policy across every copy sink. Closing this PR rather than asking you to rebase an approach that needs broader changes. If you would like to take on #939, that would be welcome.

PR #471, opened 2026-04-11
fix: use where for provider detection on Windows

### LINKED ISSUE #456: [BUG] Windows: provider detection fails - uses 'command -v' instead of 'where'
### Bug

  On Windows, zeroshot providers setup claude always reports ✗ claude CLI not found even when the claude CLI
  is installed and available in PATH.

###  Environment

  - OS: Windows 11 Pro (MINGW64 / Git Bash)
  - zeroshot: v5.4.0
  - Node: v22.17.1
  - claude: v2.1.81

###  Root Cause

  In lib/provider-detection.js, commandExists() and getCommandPath() use command -v:

 ` execSync(`command -v ${command}`, { stdio: 'pipe' });`

  command -v is a POSIX shell built-in. On Windows, Node.js execSync uses cmd.exe, which doesn't recognize it.
   The catch block silently returns false.

###  Fix

  Use where on Windows:

 ```
 const cmd = process.platform === 'win32' ? `where ${command}` : `command -v ${command}`;
  execSync(cmd, { stdio: 'pipe' });

```
  Apply in both commandExists() (line 11) and getCommandPath() (line 24).

###  Additional Note

  The ZEROSHOT_CLAUDE_COMMAND env var workaround also doesn't work from MINGW64/Git Bash because it produces
  MINGW-style paths (/c/Users/...) that where and fs.existsSync can't resolve.

###  Steps to Reproduce

  1. Install zeroshot on Windows: npm install -g @covibes/zeroshot
  2. Install Claude Code: npm install -g @anthropic-ai/claude-code
  3. Verify claude --version works
  4. Run zeroshot providers setup claude
  5. Observe ✗ claude CLI not found

haosenwang1018:
## Summary
- use `where` instead of `command -v` when detecting provider CLIs on Windows
- share the lookup command logic through a small helper used by both `commandExists()` and `getCommandPath()`
- add regression coverage for the Windows vs non-Windows command selection

## Why
Issue #456 reports that provider detection always fails on Windows because `execSync()` runs through `cmd.exe`, where `command -v` is not available.

That means `zeroshot providers setup claude` can incorrectly report `✗ claude CLI not found` even when the CLI is installed and on PATH.

## Testing
- `./node_modules/.bin/mocha tests/providers/detection.test.js`

Fixes #456

### DIFF
--- a/lib/provider-detection.js
+++ b/lib/provider-detection.js
@@ -1,6 +1,11 @@
-const { execSync, spawnSync } = require('child_process');
+const childProcess = require('child_process');
+const { execSync, spawnSync } = childProcess;
 const fs = require('fs');
 const path = require('path');
+
+function commandLookupCommand(command) {
+  return process.platform === 'win32' ? `where ${command}` : `command -v ${command}`;
+}
 
 function commandExists(command) {
   if (!command) return false;
@@ -8,7 +13,7 @@ function commandExists(command) {
     return fs.existsSync(command);
   }
   try {
-    execSync(`command -v ${command}`, { stdio: 'pipe' });
+    execSync(commandLookupCommand(command), { stdio: 'pipe' });
     return true;
   } catch {
     return false;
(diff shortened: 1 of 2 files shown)


tomdps:

Closing as superseded by #516, which merged into `dev` as `0d0ac499d3bd501e0a82b6ba9eb13aaf9b72f121`. Both PRs fix the same Windows provider-detection `command -v`/`where` bug; #516 includes the current issue reference (#515), current base branch, and regression coverage for multi-line `where` output.

Issue #457, opened 2026-03-24
Windows compatibility: npm global commands cannot be spawned without shell workaround


jmadore-payfacto:
## Summary

`zeroshot` fails on Windows when trying to spawn itself or the AI provider CLI (e.g. `claude`) because npm global commands on Windows are `.cmd` batch files that cannot be directly `spawn()`-ed by Node.js without `shell: true`. Additionally, the Jira provider uses a mix of `go-jira` and `ankitpokhrel/jira-cli` commands.

### 1. Jira provider — mixed CLI commands
- `checkAuth()` uses `jira session` (go-jira syntax)
- `fetchIssue()` uses `jira issue view --template json` (ankitpokhrel syntax)

**Fixes needed:**
- Change auth check from `jira session` → `jira me`
- Change `--template json` → `--raw` (ankitpokhrel's equivalent)

### 2. `spawnTaskProcess` in `agent-task-executor.js` — `spawn zeroshot ENOENT`
`getClaudeTasksPath()` returns `'zeroshot'` which Node.js cannot find on Windows because the actual executable is `zeroshot.cmd`.

**Fix:** On Windows, invoke `process.execPath` (node.exe) with the CLI script path directly:
```js
if (process.platform === 'win32') {
  const cliScript = path.join(__dirname, '../../cli/index.js');
  spawnCmd = process.execPath;
  spawnArgs = [cliScript, ...args];
}
```

Same issue exists in `claude-task-runner.js`.

### 3. `watcher.js` — `spawn claude ENOENT` (no task output)
`watcher.js` spawns the AI provider command (e.g. `claude`) which is also a `.cmd` file. Using `shell: true` causes arguments containing spaces (e.g. the prompt) to be split into multiple positional args, breaking the CLI invocation.

**Fix:** Resolve the actual Node.js script path from the `.cmd` wrapper file and invoke node directly. The `.cmd` file for any npm global command contains the script path and has a predictable format. Pass `commandScript` through `watcherConfig` from `runner.js` so watcher can use `spawn(process.execPath, [commandScript, ...args])` on Windows.

## Environment
- Windows 11
- Node.js (global npm install)
- `ankitpokhrel/jira-cli` for Jira

tomdps:
Status review (2026-07-29):
[...]

PR #473, opened 2026-04-11
fix: invoke zeroshot via node on Windows

haosenwang1018:
## Summary
- add a small `buildZeroshotSpawnSpec()` helper for self-invocation
- on Windows, spawn `process.execPath` with `cli/index.js` instead of trying to execute bare `zeroshot`
- reuse that helper in both `agent-task-executor` and `claude-task-runner`
- add regression coverage for the Windows vs non-Windows spawn spec

## Why
Issue #457 reports that `spawn('zeroshot', ...)` fails on Windows because npm global commands are exposed as `.cmd` wrappers, which Node cannot execute the same way as Unix binaries.

This PR fixes the smallest high-impact part first: zeroshot spawning **itself**. On Windows it now invokes Node directly with the CLI script path, avoiding the `.cmd` wrapper and the resulting `ENOENT` failures.

## Testing
- `./node_modules/.bin/mocha tests/providers/detection.test.js`

Part of #457

### DIFF
--- a/lib/provider-detection.js
+++ b/lib/provider-detection.js
@@ -1,6 +1,11 @@
-const { execSync, spawnSync } = require('child_process');
+const childProcess = require('child_process');
+const { execSync, spawnSync } = childProcess;
 const fs = require('fs');
 const path = require('path');
+
+function commandLookupCommand(command) {
+  return process.platform === 'win32' ? `where ${command}` : `command -v ${command}`;
+}
 
 function commandExists(command) {
   if (!command) return false;
@@ -8,7 +13,7 @@ function commandExists(command) {
     return fs.existsSync(command);
   }
   try {
-    execSync(`command -v ${command}`, { stdio: 'pipe' });
+    execSync(commandLookupCommand(command), { stdio: 'pipe' });
     return true;
   } catch {
     return false;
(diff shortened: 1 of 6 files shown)


tomdps:
Status review (2026-07-29):
- Windows-support theme: yes—this is the self-invocation slice of #457/#57, alongside the provider-wrapper, console-window, and IPC work in #475/#472/#63. Native Windows is currently stopped earlier by `runPreflight()`, so this defect is latent rather than a supported current path; the underlying direct spawns remain on main.
- Not merge-ready: GitHub reports merge conflicts (`mergeable=false`, `mergeable_state=dirty`) and no check runs/status contexts; the provider-detection and `windowsHide` changes are already present on main.
- Rework against current main and use one Windows-safe command/args spec for every direct host task-CLI call: the initial `task run`; launch cancellation/failure cleanup in `createPendingTaskLaunchHandle()` and `cleanupPersistedTaskAfterLaunchFailure()`; both `waitForTaskReady` variants; `_followLogs`; `createLogFollower`; `killTask`; and hook-failure log retrieval in `agent-hook-executor` (`status`, `get-log-path`, `kill`, and `logs`). The detached CLI re-exec already uses `process.execPath`; provider `.cmd` execution in `task-lib/watcher.js` remains the broader #457/#475 slice.

tomdps:
Thanks for isolating the Windows self-invocation failure. The underlying Windows executable-resolution work is still tracked in #457, but this branch predates the current task command and launch-cleanup structure, conflicts with main, and covers only part of the direct host CLI callsites. I'm closing this implementation so the remaining work can land once through a fresh, shared Windows-safe command/args abstraction rather than accumulating another partial path.

PR #476, opened 2026-04-12
fix: enforce strict schema for debug workflow tester

### LINKED ISSUE #458: tester agent JSON schema validation always fails in stream-json mode
## Summary

When using the `conductor-bootstrap` config (which triggers `debug-workflow`), the `tester` agent consistently fails with JSON schema validation errors even though the agent completes successfully and the underlying work is correct.

## Error

```
Error: Agent tester output failed JSON schema validation:
  #/required must have required property 'approved'
  #/required must have required property 'summary'
  #/required must have required property 'commandResult'
  #/required must have required property 'rootCauseVerification'
  #/required must have required property 'testVerification'
```

## Root Cause

The `tester` agent in `debug-workflow.json` runs with `stream-json` output format (for live log visibility). In this mode, zeroshot validates the `structured_output` field of the final stream-json result against the agent's output schema after the fact.

However, because `--json-schema` is silently ignored when `outputFormat !== 'json'` (see `runner.js` → `resolveJsonSchema`), Claude is never told about the required schema shape. Claude produces its standard stream-json `structured_output` with `{"summary": "...", "result": "..."}` instead of the schema-required fields (`approved`, `commandResult`, `rootCauseVerification`, `testVerification`).

The result: schema validation always fails for the tester agent, causing the cluster to reject valid work and retry indefinitely.

## Evidence

From task log, Claude's `structured_output`:
```json
{
  "summary": "Fix already applied in iteration 1; tester rejection was tester agent crash, not code issue",
  "result": "ROOT CAUSE FIX: ..."
}
```

The agent itself noted: *"The previous rejection was due to the tester agent crashing with JSON schema validation errors — the code fix was correct and already applied."*

## Expected Behavior

Either:
1. The tester agent should use `strictSchema: true` in `debug-workflow.json` so `--output-format json --json-schema <schema>` is passed to Claude, enforcing the schema natively, **or**
2. When a `jsonSchema` is defined on an agent, the system prompt should inject the schema shape so Claude knows to produce it even in `stream-json` mode

## Config Used

`conductor-bootstrap` (out of the box, no customization)

haosenwang1018:
## Summary
- enable `strictSchema: true` for the `tester` agent in `debug-workflow`
- add a template resolver regression test that locks this behavior in

## Related Issues
Fixes #458

## Changes Made
- update `cluster-templates/base-templates/debug-workflow.json` so the tester keeps native CLI schema enforcement instead of falling back to post-validation over `stream-json`
- add coverage in `tests/template-resolver.test.js` to assert the resolved `tester` agent preserves `strictSchema: true`

## Testing
- `./node_modules/.bin/mocha tests/template-resolver.test.js tests/type-safety-validation.test.js`

## Checklist
- [x] Tests pass (`npm test`)
- [ ] Documentation updated (if needed)
- [x] Follows commit guidelines

### DIFF
--- a/cluster-templates/base-templates/debug-workflow.json
+++ b/cluster-templates/base-templates/debug-workflow.json
@@ -232,6 +232,7 @@
       "modelLevel": "{{tester_level}}",
       "timeout": "{{timeout}}",
       "outputFormat": "json",
+      "strictSchema": true,
       "jsonSchema": {
         "type": "object",
         "properties": {
--- a/tests/template-resolver.test.js
+++ b/tests/template-resolver.test.js
@@ -109,6 +109,23 @@ describe('TemplateResolver', function () {
   });
 
   describe('resolve', function () {
+    it('should keep strictSchema enabled for the debug-workflow tester', function () {
+      const resolved = resolver.resolve('debug-workflow', {
+        task_type: 'DEBUG',
+        complexity: 'SIMPLE',
+        max_tokens: 100000,
+        max_iterations: DEFAULT_MAX_ITERATIONS,
+        investigator_level: 'level2',
+        fixer_level: 'level2',
+        tester_level: 'level2',
+      });
+
+      const tester = resolved.agents.find((agent) => agent.id === 'tester');
+      assert.ok(tester, 'tester agent should exist');
+      assert.strictEqual(tester.outputFormat, 'json');
+      assert.strictEqual(tester.strictSchema, true);
+    });
+
     it('should resolve single-worker template', function () {
       const resolved = resolver.resolve('single-worker', {
         task_type: 'TASK',


tomdps:
Status review (2026-07-29):
- Current main already fixes #458 generally: `validateAgentConfig`/`applyStrictSchemaDefault` defaults unset agents to strict schema while preserving the global `strictSchema=false` opt-out; `resolveOutputFormatConfig` keeps JSON and `buildTaskRunArgs` forwards `--json-schema`, while `buildFinalContext` supplies the schema for the explicit stream fallback.
- This PR is GitHub-mergeable but behind main, has no reported checks, and is not merge-ready: adding `strictSchema: true` to the `debug-workflow` tester would override that supported global opt-out, and its new test would lock in the conflicting behavior (`cluster-templates/base-templates/debug-workflow.json`; `tests/settings.test.js`).
- The linked issue is closed as fixed; this patch is superseded and should not be merged as written.

tomdps:
Thanks for chasing the debug tester schema gap. Main now fixes the underlying issue centrally: unset agents default to strict schema, while the supported global strictSchema=false opt-out is still honored. Adding an explicit true to this template would override that opt-out, so this patch is no longer the right fix. Closing as superseded.

PR #485, opened 2026-06-01
fix: surface actionable hint when Claude model unavailable on Vertex

vildursn:
## Context

When running zeroshot with `CLAUDE_CODE_USE_VERTEX=1`, the Claude CLI resolves model shorthands (`sonnet`, `haiku`) to versioned Vertex model names (e.g. `claude-sonnet-4-5@20250929`) that may not be enabled on the user's Vertex deployment. The resulting 404 error was swallowed into a raw JSON blob with no guidance on how to fix it.

## Changes

- Added `detectVertexModelError()` to parse the Claude CLI result JSON and identify Vertex model 404 errors
- `logTaskAttemptFailure` now prints an actionable fix with the exact `zeroshot settings set` command when this error is detected
- Vertex model errors short-circuit the retry loop immediately (like `HookExecutionError`) — the model being unavailable is deterministic, so retrying wastes time

## Alternatives considered

**Option B — Skip `--model` on Vertex:** When `CLAUDE_CODE_USE_VERTEX=1` is set and no `levelOverrides` are configured, pass no `--model` flag and let the CLI use the deployment default. Avoids the error entirely but surrenders model-level control.

**Option C — Inherit the parent session model:** When running under Vertex, the model the parent Claude Code session is using is guaranteed to work. Zeroshot could detect `CLAUDE_CODE_USE_VERTEX=1` and default to that model. The challenge is there's no clean way to read which model the parent session resolved to — `ANTHROPIC_MODEL` in the environment is empty even when a model is active.

🤖 Generated with [Claude Code](https://claude.ai/code)


vildursn:
Closing in favour of a clean branch based on dev

PR #496, opened 2026-06-14
fix(src/logic-engine): prevent global prototype pollution in agent sandbox

MatrixNeoKozak:
The logic engine sandbox was incorrectly attempting to prevent prototype pollution by manually freezing global constructor prototypes from the outer scope, which actually caused the reverse effect and introduced a vulnerability by modifying global behavior. This change removes the unsafe manual prototype freezing and reliance on outer-scope constructors, instead relying on `vm.createContext` to provide a genuinely isolated environment with its own standard globals. This also simplifies the sandbox execution logic and improves security.

### DIFF
--- a/src/logic-engine.js
+++ b/src/logic-engine.js
@@ -82,10 +82,6 @@ function buildAgentContext(agent) {
   };
 }
 
-function getSafeBuiltins() {
-  return { Set, Map, Array, Object, String, Number, Boolean, Math, Date, JSON };
-}
-
 function getQuietConsole() {
   return {
     log: () => {},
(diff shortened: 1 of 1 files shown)


tomdps:
Thanks for catching this.

I’m going to take this over and push an adjusted version, mainly because the current patch changes how logic scripts are evaluated. Existing Zeroshot trigger scripts are written as function bodies and rely on `return`, so we need to preserve that contract while fixing the sandbox boundary.

I’ll keep the intent of your change, add regression coverage, and make sure you’re credited for spotting the issue.

MatrixNeoKozak:
Hi @tomdps! Glad to help with finding this issue. Looking forward to your updated version in PR #521! Thanks!

PR #531, opened 2026-06-28
fix(opencode): pass --dir to opencode so agents run in the worktree, not main repo

### LINKED ISSUE #530: bug(opencode): agents operate on main repo instead of worktree due to .git pointer resolution
## Bug Description

When opencode is launched inside a git worktree, it resolves the `.git` file pointer (which contains `gitdir: /path/to/main/.git/worktrees/...`) back to the main repository root. As a result, all file operations performed by the agent target the main repo path rather than the isolated worktree.

This silently defeats worktree isolation: the agent reads and writes the wrong tree.

## Root Cause

Opencode's internal git detection follows the `.git` symlink/pointer file in the worktree directory up to the main repo. Setting the process `CWD` to the worktree path is not sufficient — opencode still resolves the true repo root.

The fix is to pass `--dir <worktree-path>` instead of (or in addition to) `--cwd`, which tells opencode explicitly which directory to treat as the working root, bypassing the gitdir pointer resolution.

## Current Behaviour

The opencode adapter detects `--cwd` support via `detectCliFeatures()` and passes `--cwd <worktree>` when available. However opencode added a `--dir` flag that prevents the gitdir walk-up; the adapter has no awareness of this flag.

## Proposed Fix

- Add `supportsDir` to `OpencodeCliFeatures` and `CliFeatureOverrides`
- Detect `--dir` in `detectCliFeatures()` via help text scan
- In `addOpencodeOptionalArgs()`, prefer `--dir` over `--cwd` when available

jregeimbal:
## Summary

When opencode is launched inside a git worktree, it follows the `.git` file pointer back to the main repository root, causing all agent file operations to target the main repo instead of the isolated worktree. Passing `--dir <worktree>` prevents this gitdir resolution.

Fixes #530.

## Motivation

Worktree isolation exists to keep each cluster's changes sandboxed. opencode's internal git detection walks `.git` (which in a worktree is a file like `gitdir: /main/.git/worktrees/...`) up to the main repo root, silently defeating the isolation — the agent reads and modifies files in the main repo rather than the worktree.

The fix is to prefer `--dir` over `--cwd` when opencode supports it. Unlike `--cwd`, `--dir` tells opencode explicitly which directory to treat as root and bypasses the gitdir pointer walk.

## Changes

- **`src/agent-cli-provider/types.ts`**: Add `supportsDir` to `OpencodeCliFeatures` and `CliFeatureOverrides`
- **`src/agent-cli-provider/adapters/opencode.ts`**: Detect `--dir` in `detectCliFeatures()` via help text scan; in `addOpencodeOptionalArgs()`, pass `--dir <cwd>` when `supportsDir` is true, falling back to `--cwd` for older opencode versions that lack the flag
- **`tests/provider-cli-builder.test.js`**: 5 new regression tests covering:
  - `--dir` is passed when `supportsDir` is true (the fix)
  - `--cwd` is used as fallback when `supportsDir` is false
  - `commandSpec.cwd` is always set regardless of CLI flag support
  - `detectCliFeatures` correctly detects `supportsDir` from help text
  - `detectCliFeatures` correctly rejects `supportsDir` when `--dir` is absent

## Testing

- [x] Unit tests added (`tests/provider-cli-builder.test.js` — 5 new tests, all passing)
- [x] Full test suite passes (1244 tests)
- [x] Linting passes (0 errors)
- [x] Type checking passes

## Breaking Changes

None. `--dir` is only passed when `detectCliFeatures()` confirms opencode supports it; older versions continue to receive `--cwd` as before.

## Checklist

- [x] Tests pass (`npm test`)
- [x] Linting passes (`npm run lint`)
- [x] Type checking passes (`npm run typecheck`)
- [x] Documentation updated (JSDoc on affected functions)
- [x] CLAUDE.md not affected (no architecture change)

### DIFF
--- a/src/agent-cli-provider/adapters/opencode.ts
+++ b/src/agent-cli-provider/adapters/opencode.ts
@@ -85,6 +85,7 @@ function detectCliFeatures(helpText?: string | null): OpencodeCliFeatures {
     supportsModel: unknown ? true : /--model\b/.test(help),
     supportsVariant: unknown ? true : /--variant\b/.test(help),
     supportsCwd: unknown ? false : /--cwd\b/.test(help),
+    supportsDir: unknown ? false : /--dir\b/.test(help),
     supportsAutoApprove: false,
     unknown,
   };
@@ -109,6 +110,8 @@ function addOpencodeOptionalArgs(args: string[], options: BuildProviderCommandOp
 
   if (options.cwd && features.supportsCwd) {
     args.push('--cwd', options.cwd);
+  } else if (options.cwd && features.supportsDir) {
+    args.push('--dir', options.cwd);
   }
 }
 
--- a/src/agent-cli-provider/types.ts
+++ b/src/agent-cli-provider/types.ts
@@ -77,6 +77,7 @@ export interface OpencodeCliFeatures extends BaseCliFeatures {
   readonly supportsModel: boolean;
   readonly supportsVariant: boolean;
   readonly supportsCwd: boolean;
+  readonly supportsDir: boolean;
   readonly supportsAutoApprove: false;
 }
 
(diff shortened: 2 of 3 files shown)


tomdps:
Recreated as #816 on a fresh branch from the new main trunk. The original commit authorship is preserved.

PR #636, opened 2026-07-11
fix(release): promote npm propagation assertion

EivMeyer:
## Summary
- promote the npm latest propagation retry fix to main
- expected semantic-release result: patch release

## Verification
- #635 pull_request CI passed
- #635 merge-group CI passed
- npm already verifies @the-open-engine/zeroshot@6.6.0 as latest; this fixes the post-release assertion false negative


EivMeyer:
Closing this promotion PR because the release artifact is already published as v6.6.0 and this branch is conflict-marked by squash-history on scripts/assert-release-published.js. The dev-only retry follow-up should be reworked separately instead of using this stale promotion PR.

Issue #640, opened 2026-07-14
Release pipeline: dev/main divergence + version drift — go trunk-based (semantic-release already in place)


EivMeyer:
## Problem

Releasing currently hand-commits to `main` (`release: promote dev to main`), which makes `main` and `dev` diverge permanently, and the `package.json` version drifts from what's on npm. Two owned-elsewhere invariants are violated: single source of truth (version) and no manual re-sync (promotion never flows back to `dev`).

**This is not a missing-tool problem.** zeroshot already runs semantic-release correctly: `.releaserc.json` releases from `main`, `@semantic-release/git` commits `chore(release): ${version} [skip ci]` back to `main`, `@semantic-release/npm` publishes, and `release.yml` runs it on CI success. The drift comes entirely from the **branch topology around it** — a long-lived `dev` branch plus a manual `dev → main` promotion flow. semantic-release commits the version bump only to `main`, that commit never flows back to `dev`, and the manual promotions pile up one-directionally.

### Evidence (as of 2026-07-13)
- `main` carries 6 commits `dev` doesn't have — **all** are release bookkeeping, no product work:
  ```
  557c7b9 release: promote dev to main (#632)
  e3111e7 release: publish dev promotion using package release config (#630)
  0e4a7ec release: publish dev promotion without branch-write release artifacts (#627)
  ec2c5d7 release: publish dev promotion with release-aware analyzer (#625)
  c242575 release: promote dev to main (#617)
  ac354df feat: release deterministic e2e and run-plan hardening (#591)
  ```
- `dev` is ~59 commits ahead with the actual work. Neither branch can fast-forward to the other.
- Both branches' `package.json` say `6.4.0`; npm `latest` is `6.6.0`. The `chore(release)` bump commit lives only on `main`, so `dev` != npm.
- **Motivating case:** #639 (fixes #637) is merged, verified, and stuck on `dev` with no clean path to release without dragging 59 unrelated commits or cherry-pick gymnastics.

## Fix: go trunk-based on `main`, delete the `dev → main` promotion flow

Align with how `the-open-engine/agents` releases (Nx Release + Conventional Commits — same philosophy, single-package equivalent here is semantic-release, which we already have). The agents repo never hits this problem **because it is trunk-based with no `dev` branch**: PRs merge straight to `main`, the release job runs on push to `main`, and semantic-release's `[skip ci]` bump commit lands on the same trunk everyone branches from — so nothing can drift.

```
[...]

Issue #637, opened 2026-07-13
Bug: `resume --detach` runs agents in the caller and leaves zombie clusters


tomdps:
> *This was generated by AI during triage.*

## Summary

`zeroshot resume <cluster> --detach` does not detach the resumed cluster into a daemon. It starts the agents inside the invoking CLI process and merely skips foreground log streaming. If that CLI process exits or is killed by a caller timeout, the cluster loses its backing process while the registry still says `running`; `zeroshot status` then reports a zombie, while another `resume` may reject it as “still running.”

## Environment

- Zeroshot: `6.3.0` (`@the-open-engine/zeroshot`)
- Node: `v24.15.0`
- OS: macOS/Darwin 25.5, arm64
- Cluster mode: `ship`
- Affected cluster: `noble-valley-29`

## Reproduction

1. Have a failed/stopped cluster with an intact worktree.
2. Run:

   ```bash
   zeroshot resume noble-valley-29 --detach "Continue the existing implementation"
   ```

3. Observe that the command does not behave like `zeroshot run --detach`: it remains the owner of the resumed agents instead of returning after a daemon handoff.
4. Let the invoking process exit or kill it with a timeout.
5. Run:

   ```bash
   zeroshot status noble-valley-29
   ```

## Observed behavior

First attempt:

```text
✓ Cluster resumed
Resume type: From failure
...
Command timed out after 180 seconds
```

After the caller killed the resume process:

```text
Cluster: noble-valley-29
State: zombie (process 88572 died, cluster has no backing process)
→ Run 'zeroshot kill noble-valley-29' to clean up, or 'zeroshot resume noble-valley-29' to restart
```

A direct retry then contradicted status:

```text
Error resuming: Cluster noble-valley-29 is still running.
Use 'zeroshot stop' first if you want to restart it.
```

The only successful workaround was:

```bash
zeroshot stop noble-valley-29
# Run resume from a separately managed long-lived/background process so the CLI stays alive.
zeroshot resume noble-valley-29 --detach "..."
```

This leaves automation responsible for daemonizing a command whose `--detach` flag promises to do so.

## Expected behavior

- `resume --detach` should perform a daemon handoff and return promptly, matching `run --detach` semantics.
- The resumed cluster must survive the parent CLI’s normal exit.
- Registry `state` and `pid` must refer to the live daemon process.
- If a stored PID is dead, `resume` should recover atomically instead of rejecting the cluster as still running.

## Root cause

[...]

PR #641, opened 2026-07-14
chore: reconcile main into dev (heal diverged release history)

EivMeyer:
## What

Reconciles the diverged `main` and `dev` histories by merging `origin/main` into `dev`. Heals the divergence root cause behind #640 and unblocks releasing the `resume --detach` zombie fix (#639, #637).

## Why

`main` and `dev` last shared history on 2026-07-09 and could not fast-forward:
- `main` carried 6 release/promotion commits (incl. `#591`, `#617`, `#632`) plus the `@semantic-release/changelog` + `@semantic-release/git` plugins in `.releaserc.json` that never flowed back to `dev`.
- `dev` carried 59 commits including the `resume --detach` fix (#639).

A plain merge hit 4 conflicts. This PR resolves them **without losing anything from either side**.

## Conflict resolution (both sides preserved)

| File | Resolution | Rationale |
|---|---|---|
| `src/orchestrator.js`, `lib/detached-startup.js` | dev | dev's shared `process-liveness` predicate supersedes main's inline pre-fix duplicates. Taking main would **revert #639**. |
| `scripts/assert-release-published.js` | dev | dev's `#635` npm-propagation-wait version supersedes main's older `#632` snapshot. |
| `tests/e2e/helpers/e2e-harness.js` | dev | dev's `buildEnv` export. |
| `.releaserc.json` | main (auto-merge) | main's semantic-release plugins are genuinely main-unique; kept. |

**Net delta vs `dev` = 8 lines in `.releaserc.json`, zero deletions vs `dev`** (fix fully intact). All of main's real feature content (`run-plan.js`, `setup-plan.js`, `release-preflight.js`, etc.) was already present on `dev` via prior promotions.

## Verification

- `npm install` + build clean.
- 34 unit tests green: fix suites (`process-liveness` 3, `detached-startup` 27) **and** main's release tooling (`release-preflight`, `assert-release-published`).
- 2 real-daemon e2e cases green: zombie recovery without manual `stop`, and single-daemon win on concurrent `resume --detach`.

## Merge instructions

**Merge as a merge commit, NOT squash.** A squash flattens the two parents and leaves `main` un-ancestored, so the branches would immediately re-diverge. The merge commit is what actually reconciles them.

Follow-up: adopt the trunk + Changesets pipeline in #640 so this drift cannot recur.


EivMeyer:
Correction on merge method: I initially wrote "merge commit, not squash." That is wrong for this repo. The `dev -> main` promotes are squashes, so `main` is never a git ancestor of `dev` by design (a squash-promote model). Reconciliation here means **content parity**, not ancestry — and the only content `dev` lacks from `main` is the 8 `.releaserc.json` lines this PR carries. So the queue's **SQUASH** method is correct: it lands those lines on `dev`, after which `dev` is a content superset of `main` and the next promote ships the #639 fix cleanly.

PR #660, opened 2026-07-15
ci: enforce Conventional Commits with a commitlint gate

EivMeyer:
## What

Adds a **commitlint gate** so the Conventional Commit messages that `semantic-release` relies on are actually enforced.

- `.commitlintrc.json` — extends `@commitlint/config-conventional`; long body/footer lines allowed (squash-merge bodies have Markdown tables and links). Mirrors the config in `the-open-engine/agents`, per "same approach everywhere."
- `.husky/commit-msg` — local validation hook (none existed; only `pre-commit`/`pre-push` were present).
- CI `commitlint` job — validates every PR commit's message via `npx`, running **before** the expensive `check` job so bad messages fail fast. Dependency-light (no `npm ci` needed for the gate).

## Why

`semantic-release` derives the version bump from commit-message *type* (`fix:` → patch, `feat:` → minor, etc.). Nothing validated those messages, so a non-conventional subject silently produces no bump or the wrong one — a contributor to the version drift tracked in #640.

## Validation

Ran the exact CI config against real messages:

```
FAIL  | updated stuff              -> subject may not be empty [subject-empty]
FAIL  | foo: bar                   -> type must be one of [...] [type-enum]
FAIL  | Feat: add thing            -> type must be lower-case [type-case]
PASS  | fix: correct version drift
PASS  | feat(release): enforce conventional commits
```

(Note: a doubled prefix like `feat: feat(x): ...` **passes** — commitlint reads `feat` as the type and the rest as the subject. This gate targets missing/unknown/wrong-case types, which is what breaks version computation.)

## Scope boundary

This PR is deliberately just the gate. The **trunk cutover** items from #640 ride with the `dev`/`main` reconciliation, not here:
- removing the `enforce-main-pr-source` CI job (which currently *mandates* the `dev → main` promotion that causes the divergence)
- swapping the release job's `GITHUB_TOKEN` for a repo-scoped release GitHub App
- the `main`-didn't-move guard on `release.yml`

## Follow-ups surfaced while implementing (for the reconciliation, not this PR)

1. **Dual semantic-release config.** `main` has `.releaserc.json` (with `@semantic-release/git` + `changelog`, which commit the version back), but `dev`'s `package.json` has an inline `release` block **without** the `git` plugin — so under dev's config no version-bump commit is written. Likely a direct contributor to branch ≠ npm drift. Consolidate to one config.
2. **Squash-merge nuance.** If PRs squash-merge, the message that reaches the trunk is the PR title, not these per-commit messages. If that's the flow, we should also lint the PR title. Left out to keep this focused.

Base is `dev` per the repo's current `enforce-main-pr-source` policy.


tomdps:
Superseded by the completed main-trunk release controls in #809 and the cutover in #810. The repository now uses the protected main trunk, semantic-release, immutable release tags, and the protected release environment.

PR #658, opened 2026-07-15
feat(gateway): add MiniMax compatibility

octo-patch:
Reason: add target provider/model to existing provider registry

## Summary

Adds MiniMax model compatibility to the bundled gateway configuration.

## Related Issues

None.

## Changes Made

- Register MiniMax-M3 and MiniMax-M2.7 in the gateway model catalog.
- Support selectable compatibility protocols and token limits in gateway requests.
- Document global and China endpoint configurations and cover request and configuration behavior with tests.

## Testing

- `npm ci`
- `npm run check:agent-cli-provider:ci`
- `npm run check`
- `npm test`

## Checklist

- [x] Tests pass (`npm test`)
- [x] Documentation updated (if needed)
- [x] Follows commit guidelines

PR #661, opened 2026-07-15
ci: reconcile dev/main + cut over to trunk (#640)

EivMeyer:
Reconciles the diverged `dev`/`main` release history and cuts `main` over to trunk. Closes the recurring manual re-sync (#633, #641) that #640 documents.

## ⚠️ Merge requires a protection toggle (see bottom) — do not "Update branch"/squash

This branch carries dev's real history (incl. its 15 merge commits), so `main`'s `required_linear_history` blocks a normal merge. Land it via the fast-forward sequence below, not the green button.

## What this is

- **`main` is now an ancestor** of this branch — divergence healed, cleanly fast-forwardable.
- **Content is byte-identical to `dev`.** Verified: `git diff <branch> origin/dev` over the reconciled tree is empty. The merge dropped only main's *stale* lines (all cosmetic — e.g. `AGENTS.md` table whitespace); nothing on `main` was real work.
- **Net change vs `dev` is only the cutover** (top commit): `ci.yml` drops `enforce-main-pr-source` + triggers on `main` only; `package.json` version → `6.6.0` (matches npm `latest`).

The huge file/commit count is just dev's accumulated work being promoted to a stale `main`. **The only thing to review is the top commit** (`ci: retire dev->main promotion…`) plus the fact that everything else equals `dev`.

## Deliberately NOT in this PR

The inline `package.json` `"release"` block stays. Removing it activates `.releaserc.json`'s `@semantic-release/git` write-back (the real fix for the version drift), which pushes to protected `main` and needs the **release GitHub App** bypass token. That lands as a follow-up once the App exists — runbook in #640.

## Landing sequence (admin)

1. Temporarily relax `main` protection: disable `required_linear_history` + `enforce_admins` (or add a bypass actor).
2. Fast-forward `main` → `reconcile/trunk-cutover` (clean; main is an ancestor).
3. Re-enable protection with the trunk config — required checks `check` + `install-matrix (…)`, add `commitlint`; `enforce-main-pr-source` no longer exists.
4. Retire `dev` (delete or ff-only); re-point open PRs (#658, #660, dependabot) to `main`; post a #tech heads-up.
5. Follow-ups: create the release App → remove the inline `release` block + wire `release.yml` → merge #660.

Refs #640.


tomdps:
Superseded by the completed main-trunk release controls in #809 and the cutover in #810. The repository now uses the protected main trunk, semantic-release, immutable release tags, and the protected release environment.

PR #737, opened 2026-07-18
fix(resume): ignore recovered ledger failures

tomdps:
## Summary

- ignore historical `AGENT_ERROR` records when the same agent has durable later task progress
- scan complete error history newest-first and select the first unresolved failure
- preserve the failed-agent resume path for genuinely unresolved ledger failures
- recover interrupted partial validation by resuming only the missing validator before aggregate rejection wakes the worker once
- cover equivalent foreground and detached resume behavior

## Root cause

`_resolveFailureInfo()` treated the newest ledger `AGENT_ERROR` as current whenever registry `failureInfo` was empty. In the reproduced cluster, that error belonged to worker iteration 1 and the worker subsequently completed iterations 2-11. The stale error bypassed clean reconstruction and resumed the worker instead of the interrupted requirements validator.

Independent review then identified a second ordering hazard: returning null as soon as the newest error was found recovered could hide an older unresolved failure. The resolver now reads complete `AGENT_ERROR` history, skips recovered candidates newest-first, and returns the newest unresolved candidate.

The lifecycle split between `TASK_STARTED` and later `TASK_ID_ASSIGNED` is expected and was not the cause.

## Evidence

- original regression first: `0 passing, 1 failing`, expected clean resume but observed failure resume
- reviewer regression first: `0 passing, 1 failing`, newest recovered error hid older unresolved validator failure
- ordering regression covers two unresolved validators followed by 12 recovered worker errors, proving newest-unresolved selection and removing the old 10-error bound
- focused recovery regressions: 3 passing
- complete `resume()` suite: 17 passing
- `npm run test:unit`: 1687 passing, 18 pending
- `npm run test:e2e`: 24 passing (isolated rerun; first run had a teardown-only ENOTEMPTY race after assertions)
- `npm run lint`: pass with existing warnings
- `npm run typecheck`: pass
- `npm run protocol:check`: pass
- `opcore check --changed --json`: passed
- earlier unchanged gates on this PR: `npm run test:slow`, `npm run rust:check`

Closes #734

PR #740, opened 2026-07-18
fix(resume): ignore recovered registry failures

### LINKED ISSUE #739: fix(resume): ignore recovered persisted registry failure during reconstruction
## Summary

`zeroshot resume` can still select a stale failed-agent path when `clusters.json` retains `failureInfo` for an agent that later completed durable work. PR #737 filters recovered ledger-derived `AGENT_ERROR` candidates, but `_resolveFailureInfo()` returns registry `cluster.failureInfo` unconditionally whenever it contains `agentId`.

Observed read-only in stopped cluster `ruby-nebula-22`:

- persisted `failureInfo`: `validator-requirements`, iteration 7, task `ancient-prism-48`, timestamp `1784204301600`;
- later durable requirements-validator completions and approved `VALIDATION_RESULT` records exist through timestamp `1784230561495`;
- `validator-code` iteration 5 has `TASK_STARTED`, `TASK_ID_ASSIGNED`, and `PROCESS_SPAWNED`, but no matching `TASK_COMPLETED`;
- current registry state still carries the old requirements-validator failure.

A resume would therefore bypass the clean planner and resume `validator-requirements`, even though complete durable history proves that failure recovered and the interrupted code validator is the correct recovery target.

## Root cause

`Orchestrator._resolveFailureInfo()` applies durable-progress filtering only to ledger-derived failures. Registry-derived failure metadata is trusted before querying lifecycle history.

## Expected behavior

- Validate persisted registry `failureInfo` against complete per-agent lifecycle history using the same recovery semantics as ledger candidates.
- If later durable task progress resolved the persisted failure, ignore it and continue checking ledger candidates before clean reconstruction.
- Preserve the failed-agent path for genuinely unresolved registry failures.
- In partial validation, clean reconstruction resumes only the interrupted validator exactly once; completed validator results remain singletons and aggregate validation behavior is unchanged.
- Planning remains read-only and impossible reconstruction starts no agents and appends no ledger events.

## Acceptance criteria

- Regression first reproduces a recovered persisted registry failure plus a different interrupted validator.
- Resume selects the interrupted validator, not the recovered registry agent.
- A truly unresolved persisted registry failure still resumes its failed agent.
- Foreground and detached recovery effects remain equivalent.
- Focused resume tests and repository quality gates pass.

tomdps:
## Summary
- ignore stale persisted registry `failureInfo` after the same agent has durable later task progress
- preserve the existing failed-agent resume path when no recovery evidence exists
- exercise recovered registry failures through both foreground and detached partial-validation recovery

## Root cause
PR #737 made ledger-derived `AGENT_ERROR` recovery durable-history aware, but `_resolveFailureInfo()` still returned persisted registry `cluster.failureInfo` unconditionally. A cluster could therefore re-enter an already recovered validator failure instead of planning the genuinely interrupted validator from the ledger.

## Regression evidence
- Red before fix: `npx mocha tests/orchestrator.test.js --grep "ignores a recovered persisted failure" --timeout 15000` -> 0 passing, 1 failing (`failure` instead of `clean`)
- Focused resume: 18 passing
- Foreground/detached partial validation: 2 passing
- Full unit: 1,694 passing, 18 pending
- Full E2E with short macOS temp root: 24 passing
- Slow/integration: 143 passing, 18 pending
- `npm run lint`: 0 errors (155 existing warnings)
- `npm run typecheck`: passed
- `npm run protocol:check`: passed
- `npm run rust:check`: passed
- `opcore check --changed`: passed

## Environment note
The first macOS E2E invocation inherited a long `$TMPDIR` and hit Unix-domain socket `listen EINVAL` plus missing fake-provider artifacts. Re-running the unchanged suite with `TMPDIR=/tmp` removed the path-length confounder and passed 24/24. The earlier Node 24 `node-pty` spawn failures were caused by installing with `npm ci --ignore-scripts`; running the repository postinstall permission repair restored the executable spawn helper, after which affected tests and the full unit suite passed.

Closes #739

### DIFF
--- a/src/orchestrator.js
+++ b/src/orchestrator.js
@@ -2649,8 +2649,14 @@ class Orchestrator {
   }
 
   _resolveFailureInfo(cluster, clusterId) {
-    if (cluster.failureInfo) {
-      return cluster.failureInfo.agentId ? cluster.failureInfo : null;
+    const persistedFailure = cluster.failureInfo?.agentId ? cluster.failureInfo : null;
+    if (persistedFailure) {
+      if (!this._hasDurableProgressAfterFailure(cluster, clusterId, persistedFailure)) {
+        return persistedFailure;
+      }
+      this._log(
+        `[Orchestrator] Ignoring recovered registry failure from ${persistedFailure.agentId}; durable task progress followed it`
+      );
     }
 
     const errors = cluster.messageBus.query({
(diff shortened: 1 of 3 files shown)


tomdps:
## Independent re-review — APPROVED

Direct inspection confirms recovered persisted `failureInfo` uses the same complete per-agent lifecycle predicate as ledger candidates; recovered registry + recovered ledger rows continue to the unresolved ledger failure, and the genuine persisted unresolved case retains failed-agent recovery. Partial-validation singleton/worker-once and fail-closed planning remain intact. Extracted helpers/callbacks comply with repository size/single-responsibility rules.

Independent proof: focused resume 19/19 and foreground/detached recovery 2/2.

Issue #718, opened 2026-07-17
bug: --model override makes conductor-loaded configs fail modelLevel validation


tomdps:
## Summary

A detached `--ship` launch with a valid provider/model override starts the conductor successfully, but the conductor's dynamic `load_config` operation is rejected because Zeroshot rewrites bootstrap agents to use `model: <override>` and then validates that generated configuration as though authored configs must use only `modelLevel`.

This stops the cluster before planner/worker execution even though preflight passed and the requested model is valid in the installed provider catalog.

## Environment

- Zeroshot: `6.7.0`
- Provider: `codex`
- Model: `gpt-5.6-sol` (locally catalogued and successfully invoked by the conductor)
- Command: `zeroshot run 669 --ship --pr-base dev --merge-queue --provider codex --model gpt-5.6-sol -d`
- Cluster: `jade-tower-80`
- Run mode: detached ship

## Reproduction evidence

Preflight and the conductor task succeeded:

```text
✓ Preflight checks passed
Started jade-tower-80 (ship)
junior-conductor | ⚡ ISSUE_OPENED → task #1 (gpt-5.6-sol)
junior-conductor | ✓ completed
```

The conductor returned a valid STANDARD/TASK classification, but the dynamic operation failed:

```text
Failed to execute CLUSTER_OPERATIONS: Proposed cluster configuration is invalid:
  - Agent 'junior-conductor' uses 'model: "gpt-5.6-sol"'. Use 'modelLevel: "level1|level2|level3"' instead for provider-agnostic model selection.
  - Agent 'senior-conductor' uses 'model: "gpt-5.6-sol"'. Use 'modelLevel: "level1|level2|level3"' instead for provider-agnostic model selection.
  - Agent 'git-pusher' uses 'model: "gpt-5.6-sol"'. Use 'modelLevel: "level1|level2|level3"' instead for provider-agnostic model selection.
```

The cluster then became `stopped` with `failureInfo: null`; the senior conductor retained `currentTask: true` in the registry.

Relaunching the same issue with `--provider codex` but without `--model` allowed the dynamic config to load and planner execution to begin.

## Expected

CLI `--model` is an explicit provider-specific escape hatch. It must remain authoritative for bootstrap and dynamically loaded agents without making the internally materialized configuration invalid.

If an override cannot be applied, preflight must fail before starting agents with a typed actionable error.

## Impact

- Valid explicit model overrides cannot run normal conductor workflows.
[...]

PR #792, opened 2026-07-24
fix: preserve CLI model override across config loads

shixi-li:
## Summary

Preserve a validated CLI model override when a conductor dynamically loads another cluster config. Existing agent configs now drop only their runtime-materialized `model` projection before operation-chain validation, while newly authored raw `model` fields remain subject to the repository's `modelLevel` policy.

## Related Issues

Addresses https://github.com/the-open-engine/zeroshot/issues/718

## Changes Made

- Clone and normalize already-persisted agent configs before validating a proposed dynamic config.
- Make same-ID proposal replacement match runtime replacement semantics, keeping add/load/update payloads visible to validation.
- Reject authored raw model fields even when their value equals the active override.
- Cover the conductor `load_config` path in-process and through a real detached CLI daemon.
- Route validation, execution, and topology simulation through one config-reference resolver.
- Reject path-like static config and base-template names before filesystem access.

## Scope Boundary

This PR fixes successful CLI model-override propagation through conductor `load_config`. It does not change the generic `failureInfo` persistence or agent task/PID reconciliation behavior also described in #718; those lifecycle-hardening items remain follow-up work.

## Testing

- `node tests/run-tests.js tests/model-override.test.js` (11 passing)
- `node tests/run-tests.js tests/e2e/cluster-lifecycle.test.js` (2 passing)
- `npm run lint`
- `npm run typecheck`
- `npm test` (1708 passing, 18 pending)
- `npm run test:e2e` (25 passing)
- `npm run test:slow`
- `npm run dupcheck`
- `npm audit --audit-level=moderate --omit=dev`
- Targeted Prettier and `git diff --check origin/dev...HEAD`

`npm run format:check` still reports the same 15 pre-existing files on unmodified `dev` at `ef5ed29`; none overlaps this change, and all changed files pass targeted Prettier.

## Checklist

- [x] Tests pass (`npm test`)
- [x] Documentation updated (not needed; behavior-only fix)
- [x] Follows commit guidelines

### DIFF
--- a/src/orchestrator.js
+++ b/src/orchestrator.js
@@ -3462,7 +3462,7 @@ Continue from where you left off. Review your previous output to understand what
 
     // Phase 2: Build mock cluster config with proposed agents
     // Collect all agents that would exist after operations complete
-    const existingAgentConfigs = cluster.config.agents || [];
+    const existingAgentConfigs = this._prepareExistingAgentConfigsForValidation(cluster);
     const proposedAgentConfigs = this._buildProposedAgentConfigs(existingAgentConfigs, operations);
 
     // Phase 3: Validate proposed cluster config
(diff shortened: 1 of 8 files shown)


tomdps:
Recreated as #814 on a fresh branch from the new main trunk. The original commits and shixi-li’s authorship are preserved.

PR #804, opened 2026-07-27
fix(agent): strip null bytes from spawn args before launching task process

### LINKED ISSUE #621: Agent spawn crashes when prompt contains a null byte (prompt passed as argv)
## Summary

A ship run dies during implementation with:

```
TypeError [ERR_INVALID_ARG_VALUE]: The argument 'args[10]' must be a string without null bytes.
Received 'You are agent "worker" with role "implementation".\nIteration: 3\n## 🔴 CRITICAL: AUTONOMOUS EXECUTION REQUIRED\n...'
    at normalizeSpawnArguments (node:child_process:584:3)
    at spawn (node:child_process:780:13)
    at spawnTaskProcess (src/agent/agent-task-executor.js:833:18)
    at spawnClaudeTask (src/agent/agent-task-executor.js:654:24)
```

Version: `@the-open-engine/zeroshot@6.4.0`.

## Root cause

`spawnTaskProcess` (`src/agent/agent-task-executor.js:833`) passes the full agent prompt as a **command-line argument** (`args[10]`). Node's `child_process.spawn` rejects any argv entry containing a NUL (`\0`), so if the assembled prompt contains a null byte the whole task spawn throws and the run fails.

Two problems compound here:
1. **Prompts are passed as argv.** Prompts are large and can contain arbitrary bytes; passing them as process arguments is fragile (also hits ARG_MAX on big contexts).
2. **Binary content leaks into the prompt.** This reproduced on a repo that checks in a native binary (`opcore-graph-core`). Repository/diff context that includes binary bytes puts a `\0` into the prompt string, triggering the crash.

## Suggested fix

- Deliver the prompt via **stdin or a temp file**, not argv. This removes both the NUL restriction and the ARG_MAX ceiling.
- Independently, **strip/replace NUL bytes** (and ideally skip binary blobs) when assembling repo/diff context into a prompt.

## Impact

Any run whose context picks up binary file content fails at the implementation handoff and cannot ship. Observed on runs `bright-nova-52` (later re-run `ivory-condor-39`) against a repo containing a committed native binary.

atharvwasthere:
## Problem

A task prompt containing a null byte crashes the whole cluster. `spawnTaskProcess` passes the built args array, including the full prompt, straight to `child_process.spawn()`. Node's `normalizeSpawnArguments` throws `TypeError [ERR_INVALID_ARG_VALUE]` on any argv string with an embedded null byte, because argv strings are null terminated at the OS level and simply cannot carry one.

```
TypeError [ERR_INVALID_ARG_VALUE]: The argument 'args[10]' must be a string without null bytes.
    at normalizeSpawnArguments (node:child_process:584:3)
    at spawn (node:child_process:780:13)
    at spawnTaskProcess (src/agent/agent-task-executor.js:833:18)
```

Fixes #621

## Fix

Strip null bytes from string args right before the `spawn()` call in `spawnTaskProcess` (`src/agent/agent-task-executor.js`). This is the one place all task-run args funnel through before reaching the OS, so it covers the prompt and any other string arg without touching prompt-building code upstream.

Also exported `spawnTaskProcess` so it can be unit tested directly, same as other internals already exported from this file.

## Test plan

- Added `tests/unit/spawn-null-byte-sanitization.test.js`:
  - stubs `child_process.spawn` and asserts a prompt with an embedded null byte reaches `spawn()` with the null byte stripped
  - runs the real (unstubbed) `spawn()` with a null-byte-containing arg and asserts it does not throw `ERR_INVALID_ARG_VALUE`
- Verified the test fails with `ERR_INVALID_ARG_VALUE` when the fix is reverted, confirming it actually catches the original bug
- Full unit suite: 1212 passing, 0 new failures
- `npm run lint` / `npm run typecheck` clean on changed files (repo currently has pre-existing, unrelated lint debt on `dev` itself, not introduced by this change)

### DIFF
--- a/src/agent/agent-task-executor.js
+++ b/src/agent/agent-task-executor.js
@@ -875,9 +875,7 @@ function spawnTaskProcess({ agent, ctPath, args, cwd, spawnEnv }) {
   // Timeout for spawn phase - if CLI hangs during init (e.g., opencode 429 bug), kill it
   const SPAWN_TIMEOUT_MS = 30000; // 30 seconds to spawn task
 
-  // argv strings are null-terminated at the OS level, so a null byte anywhere
-  // in agent-generated content (e.g. the task prompt) crashes spawn() with
-  // ERR_INVALID_ARG_VALUE. Strip it rather than letting it reach spawn().
+  // spawn() throws on null bytes in argv; strip them before they get there.
   const safeArgs = args.map((arg) => (typeof arg === 'string' ? arg.replace(/\0/g, '') : arg));
 
   return new Promise((resolve, reject) => {
--- a/src/agent/agent-task-executor.js
+++ b/src/agent/agent-task-executor.js
@@ -875,8 +875,13 @@ function spawnTaskProcess({ agent, ctPath, args, cwd, spawnEnv }) {
   // Timeout for spawn phase - if CLI hangs during init (e.g., opencode 429 bug), kill it
   const SPAWN_TIMEOUT_MS = 30000; // 30 seconds to spawn task
 
+  // argv strings are null-terminated at the OS level, so a null byte anywhere
+  // in agent-generated content (e.g. the task prompt) crashes spawn() with
+  // ERR_INVALID_ARG_VALUE. Strip it rather than letting it reach spawn().
+  const safeArgs = args.map((arg) => (typeof arg === 'string' ? arg.replace(/\0/g, '') : arg));
+
   return new Promise((resolve, reject) => {
-    const proc = spawn(ctPath, args, {
+    const proc = spawn(ctPath, safeArgs, {
       cwd,
       stdio: ['ignore', 'pipe', 'pipe'],
       env: spawnEnv,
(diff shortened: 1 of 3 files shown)


atharvwasthere:
Good catch, thanks. Confirmed the same crash is reachable through `IsolationManager.spawnInContainer` in `src/isolation-manager.js`, it builds `args` from the same unsanitized prompt and calls `spawn('docker', args, ...)` directly, bypassing the sanitization I added in `spawnTaskProcess`.

Fixed in d10507f: same one-line strip applied at that call site, plus a regression test for the isolated path. Both call sites now sanitize independently since they're two separate `spawn()` entry points, not one shared function.

atharvwasthere:
@EivMeyer this fixes #621. Root cause confirmed: `spawnTaskProcess` passed the prompt straight into argv, and argv strings are null-terminated at the OS level so a null byte crashes `spawn()` with `ERR_INVALID_ARG_VALUE`. The Docker-isolated path (`IsolationManager.spawnInContainer`) had the same bug independently, fixed that too. Both call sites now strip null bytes before spawning, with regression tests for each.

tomdps:
Recreated as #813 on a fresh branch from the new main trunk. The original three commits and Atharv Singh’s authorship are preserved.

PR #806, opened 2026-07-27
fix: isolate Claude safety hooks per run

### LINKED ISSUE #43: [FEATURE] Move hook configuration to project path
## Problem

Choosing to run zeroshot in one project directory shouldn't impact other projects/directories.

When running zeroshot for the first time it immediately changes the user level `~/.claude/settings.json` which then causes the AskUserQuestion/DangerousGit hooks to be loaded/run into the claude cli in all cases. While it's clear these hooks DO check for the zeroshot ENV when they execute, they still get triggered and executed. 

## Proposed Solution

Move the hook configuration to the project level `.claude/settings.json`. 

## Alternatives Considered

Leaving the configuration as is.

tomdps:
## Summary

- stop installing Zeroshot safety hooks into the user-level Claude settings
- create a mode-0600 temporary settings overlay for each Claude task and load it through the supported --settings flag
- preserve normal user, project, local, credential, plugin, skill, and MCP discovery
- add the dangerous-git hook only for worktree tasks and clean the overlay after the provider exits
- route the overlay path through the provider command contract and cover the isolation behavior with regression tests

## Root cause

The prior project-overlay helper returned null when a repository had no Claude config. The hook installers treated that null as permission to fall back to CLAUDE_CONFIG_DIR or ~/.claude, so the first Zeroshot run mutated global Claude settings.

## Validation

- npm run check:agent-cli-provider:ci — 135 passing
- focused hook/config suite — 52 passing
- npm run check — pass (0 errors; existing warnings only)
- full npm test — 1,697 passing, 18 pending; 7 unrelated PTY tests fail in this worktree because node-pty reports posix_spawnp failed
- Python hooks compile successfully with python3 -m py_compile
- Opcore targeted validation is degraded because this repository has no configured Python type-checker authority for the two comment-only hook files; direct syntax and hook behavior checks pass

## Migration note

This prevents future global settings mutations. It deliberately does not rewrite existing user-level Claude settings to remove hook entries installed by an older Zeroshot version, because an automatic cleanup would itself mutate user-owned global configuration.

Fixes #43


tomdps:
Superseded by #812, which preserves these three commits and their authorship on a fresh branch from the new main trunk.

PR #808, opened 2026-07-27
feat: reuse provider sessions per agent

### LINKED ISSUE #482: feat: opt into provider-side prompt caching to reduce per-cluster token cost
## Context

Running cross-model clusters (Claude planner + Codex worker + 2x Claude validators) on a production codebase, cost-per-ticket lands at \$4-15. A meaningful portion comes from the planner, validators, and worker-iterations each paying full input-token cost on the same static prefix (system prompt + ISSUE_OPENED + PLAN_READY + prior context).

Anthropic and OpenAI both support prompt prefix caching, but zeroshot doesn't currently opt in. Quick grep across `covibes/zeroshot`:

- No matches for `cache_control` / `cacheControl` / "prompt cache" in the code.
- No existing issues mention prompt caching.

## Why this matters

**Anthropic** (Claude, used by planner + validators in cross-model clusters):
- Cache reads cost ~10% of normal input tokens (90% off).
- Requires explicit `cache_control: { type: "ephemeral" }` markers on the static portion of the request — automatic caching is opt-in via these markers, not default.
- 5-minute TTL — long enough to span planner → validator on the same ticket.

**OpenAI** (Codex worker):
- Automatic prefix caching for prompts >1024 tokens at 50% discount on cached prefix.
- Triggered by stable prefix ordering — no API opt-in needed, but the prompt structure must be deterministic.

**Cluster math**: planner runs ONCE (~10-30k input tokens), then both Claude validators read PLAN_READY + ISSUE_OPENED + IMPLEMENTATION_READY in their context. If the planner's static prefix is byte-identical with each validator's static prefix (system prompt + ISSUE_OPENED block in the same position), the validators hit the cache and pay ~10% on that portion. On a 5-iteration cluster that's roughly 30-40% off the validator stack.

## Concrete ask

1. **Anthropic provider**: insert `cache_control: { type: "ephemeral" }` on the last content block of the system prompt and on the static portion of the user/assistant message containing ISSUE_OPENED + PLAN_READY (the parts that don't change across agents). The change is local to the Anthropic request builder.

2. **OpenAI / Codex provider**: ensure the static portion of the prompt (system + ISSUE_OPENED + PLAN_READY) appears at the very beginning of the prompt, in deterministic order, before any per-iteration content. Likely already true given `contextStrategy.format: "chronological"`, but worth verifying.

3. **All providers**: document the resulting cache-hit behavior in the cluster docs so config authors know which portions of `contextStrategy.sources` benefit from being earlier in the chronological order.

## Acceptance criteria

- AC1 (MUST): Anthropic provider sets `cache_control` markers such that a real cluster run on a representative ticket reports cache hits on validator and worker iterations. Verify via the API response's `usage.cache_read_input_tokens` / `cache_creation_input_tokens` fields.
- AC2 (MUST): OpenAI provider doesn't regress; the static-prefix ordering is verified to be deterministic via a unit test on the request builder output.
- AC3 (SHOULD): Telemetry — either the cluster summary or the per-agent log emits cache-read / cache-creation token counts so operators can measure the win.
- AC4 (NICE): Docs section in README explaining the caching behavior and how `contextStrategy` source ordering affects it.

## Out of scope

- Cross-agent shared context (a sibling problem — see follow-up scratchpad issue when filed).
- Cache markers on the worker's per-iteration changing content (would be wrong; only the static prefix should be cached).
- Cache invalidation on file changes (provider TTL handles this implicitly at 5 min).

Happy to PR if helpful — would want a quick read from a maintainer first on whether the Anthropic provider builder is the right place to inject `cache_control`, or whether it should live in the cluster's message-assembly layer.

tomdps:
## Summary

- capture Claude session IDs and Codex thread IDs from provider JSONL output
- persist continuation state on the owning logical agent and restore it with cluster state
- resume only that same agent by explicit session ID; never use cwd-wide latest-session selection
- fail closed to a fresh session on provider changes, failed tasks, unsupported CLI versions, or Docker isolation
- make manual task resume use the captured explicit provider session for Claude and Codex

This follows the revised scope in #482: provider-side session reuse rather than direct API `cache_control` plumbing.

## Validation

- `npm run check`
- `npm run check:agent-cli-provider:ci` (135 passing)
- focused provider/session/persistence tests (24 passing)
- `opcore check --changed --json`
- `npm test`: 1705 passing, 18 pending; 7 local attach/PTTY tests fail with the environment-level `node-pty` error `posix_spawnp failed`

Fixes #482

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,6 +49,7 @@ Destructive commands (need permission): `zeroshot kill`, `zeroshot clear`, `zero
 | Gateway tools/policy         | `src/agent-cli-provider/gateway-tools.ts`                               |
 | Provider detection           | `lib/provider-detection.js`                                             |
 | Provider capabilities        | `src/providers/capabilities.js`                                         |
+| Provider session reuse       | `src/agent/provider-session.js`                                         |
 | Start-cluster helper         | `lib/start-cluster.js`                                                  |
 | Legacy worker facade         | `lib/cluster-worker/`                                                   |
 | Legacy worker executable     | `bin/zeroshot-cluster-worker.js`                                        |
@@ -331,6 +332,12 @@ Restart persistence: orchestrator publishes `AGENT_RESTART_ATTEMPT` to the ledge
 Provider task ownership: task watchers persist an owned termination boundary with each active task.
 POSIX providers run in a dedicated process group; Windows providers use the exact root PID with
 `taskkill /T`. Recovery must terminate that recorded boundary before retrying work.
+
+Provider session reuse is explicit-ID and agent-owned. Watchers may capture provider session IDs,
+but only the same logical `AgentWrapper` may resume them. Persist this state in that agent's
+`agentStates` entry, never select a cwd-wide "latest" session, never share across agents, and start
+fresh when provider capability, identity, task completion, or durable isolation semantics do not
+prove reuse is safe.
 
 ### Guidance Messaging
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -333,11 +333,15 @@ Provider task ownership: task watchers persist an owned termination boundary wit
 POSIX providers run in a dedicated process group; Windows providers use the exact root PID with
 `taskkill /T`. Recovery must terminate that recorded boundary before retrying work.
 
-Provider session reuse is explicit-ID and agent-owned. Watchers may capture provider session IDs,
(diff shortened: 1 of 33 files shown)


tomdps:
Recreated on a fresh `main` base as #811. The migrated branch preserves the original isolated diff and commit author while removing legacy `dev` history.

PR #823, opened 2026-07-27
fix: fail closed on interactive Keychain use in darwin workers

### LINKED ISSUE #704: bug: non-interactive worktree clusters can launch macOS Keychain UI

## Bug

A local `--worktree`/`--ship` cluster can repeatedly launch macOS `SecurityAgent` dialogs from commands run by non-interactive workers. The agent is declared non-interactive, but descendant processes still have access to the logged-in user's GUI Keychain session.

This is not a timing-only correlation. A live process trace captured the exact chain while the popup was visible:

```text
security -i
└─ claude ... doctor
   └─ pytest -q
      └─ uv run --locked --no-sync pytest -q
         └─ codex exec ... -C ~/.zeroshot/worktrees/bronze-forest-77 ...
            (agent: validator-requirements, cluster: bronze-forest-77)
```

At 2026-07-16 15:05:44 PDT, the user reported the popup "right now". The simultaneously captured processes were:

```text
16516 16465 security -i
16465 16094 ~/.local/bin/claude --bare --settings .../test_real_pinned_cli_accepts_e0/profile/settings.json ... doctor
16094 16090 ~/.zeroshot/worktrees/bronze-forest-77/.venv/bin/python .../pytest -q
16090  7635 uv run --locked --no-sync pytest -q
 7635  7632 .../codex exec ... -C ~/.zeroshot/worktrees/bronze-forest-77 ...
```

The cluster was validating `the-open-engine/benchmarks#131`. Its ordinary suite includes `tests/test_claude_profile.py::test_real_pinned_cli_accepts_exact_offline_profile_when_installed`, which runs the real installed Claude CLI. `inspect_claude_profile()` calls `claude ... doctor`.

A PATH probe substituted a non-mutating fake `security` executable and captured Claude Doctor's exact request:

```text
argv: <-i>
stdin: add-generic-password -U -a "unknown" -s "Claude Code-doctor-probe" -X "70726f6265"
```

That is the popup source: Claude Doctor deliberately probes Keychain writes through `security -i`. Every full-suite validation iteration repeats the probe, so multi-agent/retry cycles make the dialogs appear continuously.

## Deterministic reproduction

On macOS with the pinned Claude CLI installed:

```bash
cd ~/.zeroshot/worktrees/bronze-forest-77
UV_PYTHON=3.12.13 uv run --locked --no-sync pytest -q \
  tests/test_claude_profile.py::test_real_pinned_cli_accepts_exact_offline_profile_when_installed
```

Observed: `1 passed in 2.99s`, while the process trace captured `pytest -> claude doctor -> security -i` and macOS launched `SecurityAgent`.

The same targeted test was then rerun with a fake `security` first on `PATH`; no dialog appeared, the test still passed, and the captured stdin above proved the write probe.

## Zeroshot path that permits it

- `src/agent/agent-task-executor.js:779-820` builds the local worker environment from `{ ...process.env }`.
- `task-lib/watcher.js:40-48` again inherits the environment and spawns the provider CLI.
- Worktree mode isolates Git state, but it does not isolate descendant commands from the host GUI Keychain session.

This is distinct from #544/#553. Those changes provision the provider's own Claude credentials. Here, a Codex worker launches a repository command, that command launches a second credential-aware CLI, and that grandchild opens GUI authorization dialogs during a supposedly non-interactive cluster.

## Expected behavior

Non-interactive clusters must not surface interactive OS credential dialogs from worker descendants. A credential/keychain attempt should either:

1. fail closed with a diagnostic that identifies the command and recommends Docker/explicit credential configuration; or
2. run behind a credential boundary that preserves the provider's own authentication without exposing the user's interactive Keychain session to arbitrary worker descendants.

Do not silently break provider authentication or merely suppress the visible dialog while allowing the operation.

## Acceptance criteria

- A macOS local/worktree worker descendant that attempts the equivalent of `security -i` cannot launch `SecurityAgent`.
- The worker receives a deterministic non-interactive failure or uses an explicitly configured credential path.
- Zeroshot's own provider authentication still works for Claude/Codex agents.
- Docker behavior remains unchanged.
- A regression test covers a worker grandchild attempting an interactive Keychain operation without modifying a developer's real Keychain.

## Environment

- macOS 26 / Darwin 25.5.0, arm64
- Zeroshot 6.6-series installed CLI
- local `--ship` cluster using worktree isolation
- Codex worker; descendant Claude Code `doctor`

shixi-li:
## Problem

Non-interactive local/worktree worker clusters on macOS can pop GUI Keychain password prompts (SecurityAgent). `buildSpawnEnv` spawns workers with the inherited host environment, and `task-lib/watcher.js` passes it through to every worker grandchild, so a descendant like `claude doctor` probing Keychain writes via `security -i` reaches the logged-in user's GUI Keychain session.

Fixes https://github.com/the-open-engine/zeroshot/issues/704

## Fix (fail closed, option 1 from the issue)

On darwin only, worker spawn envs now get a managed shim directory (`~/.zeroshot/keychain-shim`) prepended to the exact `PATH` key. Its `security` wrapper:

- fails closed (exit 1) on interactive invocations — `-i`, `-p` (implies interactive), bundled global flags like `-qi`, or a zero-argument implicit interactive session — with a deterministic diagnostic naming the non-interactive worker boundary and pointing at Docker isolation or explicit credential configuration;
- transparently `exec`s `/usr/bin/security "$@"` for every other subcommand, so provider authentication such as `security find-generic-password` keeps working (flag scanning stops at the first subcommand token, and exit codes pass through).

`ZEROSHOT_ALLOW_INTERACTIVE_KEYCHAIN=1` opts out. If the shim cannot be installed, the spawn fails rather than silently re-exposing the interactive Keychain session.

The managed executable is refreshed through a same-directory temporary file, chmodded before an atomic rename over the live path, and cleaned up on publication failure. PATH handling preserves inherited empty components (the POSIX current-directory entry), distinguishes unset PATH from an explicitly empty PATH, and removes only duplicate shim entries.

Scope guarantees:

- Docker isolation never reaches `buildSpawnEnv`, so containerized runs are untouched.
- Non-darwin platforms get no shim.
- Worktree tool-bin directories stay ahead of the shim, so repo-managed tool substitutes keep winning.

## Tests

- `tests/unit/darwin-keychain-boundary.test.js` now has 15 cases covering interactive blocking, non-interactive pass-through and exit propagation, opt-out behavior, exact `PATH` casing when `Path` also exists, empty/unset PATH semantics, idempotency, atomic publication failure, and non-darwin behavior.
- Both production spawn-env builders expose a narrow boundary-injection seam. Their tests force the darwin path on every CI platform, use real worktree tooling metadata, and launch a worker process whose PATH-resolved `security -i` grandchild must hit the shim rather than a logging fallback executable.
- Focused boundary + adjacent worktree/runner tests: 24 passing.
- Full unit suite, lint, and typecheck complete successfully.

## Notes

A PATH shim intercepts PATH-resolved lookups; descendants invoking `/usr/bin/security` by absolute path bypass it. The reported probe resolves via PATH, so the defect in the issue is covered; this is a guardrail for well-behaved tools, not a sandbox against adversarial code.

### DIFF
--- a/src/agent/agent-task-executor.js
+++ b/src/agent/agent-task-executor.js
@@ -19,6 +19,7 @@ const { getTask, getTaskBySpawnOwnershipToken } = require('../../task-lib/store.
 const { loadSettings } = require('../../lib/settings.js');
 const { resolveClaudeAuth } = require('../../lib/settings/claude-auth.js');
 const { prependWorktreeToolBinToEnv } = require('../worktree-tooling-env.js');
+const { applyDarwinKeychainBoundaryToEnv } = require('../darwin-keychain-boundary.js');
 const {
   CLAUDE_SETTINGS_ENV,
   cleanupClaudeSettingsOverlay,
@@ -723,6 +724,13 @@ function buildSpawnEnv(agent, providerName, modelSpec, options = {}) {
       spawnEnv.ZEROSHOT_WORKTREE = '1';
     }
   }
+
+  // KEYCHAIN BOUNDARY (darwin only): non-interactive local/worktree worker
+  // descendants must not reach the user's GUI Keychain session (issue #704).
+  // Docker isolation never reaches buildSpawnEnv (see spawnClaudeTaskIsolated).
+  // Applied before the worktree tool bins so repo-managed tool substitutes
+  // stay first on PATH.
+  applyDarwinKeychainBoundaryToEnv(spawnEnv);
 
   prependWorktreeToolBinToEnv(spawnEnv, {
     cwd: agentCwd,
@@ -2628,6 +2636,7 @@ module.exports = {
   ensureAskUserQuestionHook,
   ensureDangerousGitHook,
   resolveMcpConfigArgs,
+  buildSpawnEnv,
   spawnClaudeTask,
   spawnTaskProcess,
   followClaudeTaskLogs,
--- a/src/agent/agent-task-executor.js
+++ b/src/agent/agent-task-executor.js
@@ -690,7 +690,10 @@ function buildFinalContext({ agent, context, desiredOutputFormat, runOutputForma
 }
 
 function buildSpawnEnv(agent, providerName, modelSpec, options = {}) {
-  const { claudeSettingsPath = null } = options;
+  const {
+    claudeSettingsPath = null,
+    applyDarwinKeychainBoundary = applyDarwinKeychainBoundaryToEnv,
+  } = options;
   const spawnEnv = { ...process.env };
   const agentCwd = agent.config?.cwd || agent.worktree?.path || process.cwd();
   const clusterId = agent.cluster?.id || agent.cluster_id || process.env.ZEROSHOT_CLUSTER_ID;
@@ -730,7 +733,7 @@ function buildSpawnEnv(agent, providerName, modelSpec, options = {}) {
   // Docker isolation never reaches buildSpawnEnv (see spawnClaudeTaskIsolated).
   // Applied before the worktree tool bins so repo-managed tool substitutes
   // stay first on PATH.
-  applyDarwinKeychainBoundaryToEnv(spawnEnv);
+  applyDarwinKeychainBoundary(spawnEnv);
 
   prependWorktreeToolBinToEnv(spawnEnv, {
     cwd: agentCwd,
@@ -830,14 +833,7 @@ function createPendingTaskLaunchHandle({
   return handle;
 }
 
-function spawnTaskProcess({
-  agent,
-  ctPath,
-  args,
-  cwd,
-  spawnEnv,
-  spawnTimeoutMs = 30000,
-}) {
+function spawnTaskProcess({ agent, ctPath, args, cwd, spawnEnv, spawnTimeoutMs = 30000 }) {
   // Timeout for spawn phase - if CLI hangs during init (e.g., opencode 429 bug), kill it.
   const SPAWN_TIMEOUT_MS = spawnTimeoutMs;
   // spawn() throws on null bytes in argv; strip them before they get there.
@@ -886,10 +882,7 @@ function spawnTaskProcess({
     const classifyCleanupOwnership = trackTaskWrapperCleanupOwnership(findPersistedTaskId);
     const rejectWithOwnership = async (error) => {
       const classifiedError = classifyCleanupOwnership(error);
-      if (
-        !callerOwnsCommandCleanup(classifiedError) &&
-        agent.currentTask === pendingLaunch
-      ) {
+      if (!callerOwnsCommandCleanup(classifiedError) && agent.currentTask === pendingLaunch) {
         let termination;
         try {
           termination = await pendingLaunch.kill(classifiedError.message);
@@ -2205,7 +2198,6 @@ function startIsolatedTail({ agent, manager, clusterId, logFilePath, state, onLi
     agent._log(`[${agent.id}] tail process error: ${err.message}`);
   });
 }
-
 
 async function checkIsolatedStatus({
   agent,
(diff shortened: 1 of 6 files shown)


tomdps (changes requested):
The reported #704 root cause is real, but this boundary is not ready to merge. Please address these concrete issues:

1. `src/darwin-keychain-boundary.js:27`: Darwin environment keys are case-sensitive. `pathKeyForEnv()` can select `Path` when both `Path` and `PATH` exist, leaving the actual `PATH` used by descendants unchanged and allowing `security -i` to bypass the shim. On Darwin, update the exact `PATH` key and add a both-keys regression case.

2. `src/darwin-keychain-boundary.js:98-102`: publish the executable atomically. `writeFileSync()` truncates the live shim, so concurrent installers or a failed refresh can expose an empty/partial `security` executable to existing workers. Write and chmod a same-directory temporary file, then rename it over the live path; clean the temporary file on failure.

3. `src/darwin-keychain-boundary.js:135-146`: preserve inherited empty PATH components and distinguish unset PATH from explicitly empty PATH. Empty POSIX PATH segments mean the current directory; filter only duplicate shim-directory entries before prepending the shim.

4. `tests/unit/darwin-keychain-boundary.test.js:217-218`: the production-wiring cases take only the non-Darwin branch in required Ubuntu unit CI, while the required macOS install matrix does not run this test. Inject/stub the platform boundary in both production builders and exercise a PATH-resolved `security -i` grandchild using the returned environment, including real worktree tool-bin ordering. The test must fail if either production `applyDarwinKeychainBoundaryToEnv` call is removed.

Please also split `tests/e2e/helpers/e2e-harness.js` cleanup retry into a separate PR; it is reasonable but unrelated to the Keychain boundary squash.

After those updates, rerun the required gates and request re-review.

shixi-li:
@tomdps Thanks for the concrete review. I rebased the change onto current `main` and addressed every requested item in `c892e4a`:

1. Darwin now updates the exact `PATH` key. A regression case starts with both `Path` and `PATH` and verifies only the real `PATH` is changed.
2. Shim refreshes write and chmod a uniquely named same-directory temporary file, atomically rename it over `security`, and remove the temporary file on failure. The failure-path test forces `renameSync` to fail and verifies the existing live shim remains intact with no temporary file left behind.
3. PATH mutation now removes only duplicate shim-directory entries. Empty components are retained, and tests separately cover unset PATH (`<shim>`) and explicitly empty PATH (`<shim>:`).
4. Both production spawn-env builders now have a narrow boundary-injection seam. Their tests force the Darwin branch on every CI platform, create real worktree tooling metadata, assert `tool-bin -> shim -> inherited PATH` ordering, and launch a worker whose PATH-resolved `security -i` grandchild must be blocked by the shim. Removing either production boundary call makes its corresponding test reach the logging fallback executable and fail.
5. The unrelated `tests/e2e/helpers/e2e-harness.js` cleanup retry is no longer present in this PR's commits or diff.

Validation after the rebase:

- focused boundary + adjacent worktree/runner tests: 24 passing
- complete `npm run test:unit`: exit 0
- `npm run lint`: exit 0 (existing warnings only)
- `npm run typecheck`: exit 0
- pre-push lint and typecheck gates: exit 0

Could you please re-review when convenient?

tomdps (approved):
Re-reviewed exact head 936d2eb6f32dc02a8753206eb00850ebc2c7ab85 after the contributor updates and two follow-up reviewer-fix commits. The #704 root cause is confirmed, every prior finding is resolved, the full current diff and direct spawn/PATH/cleanup integrations are clean, regression coverage exercises both production builders, unrelated cleanup is absent, authorship is preserved, and all applicable required gates are green. Approved for squash merge at this unchanged head.

Issue #766, opened 2026-07-23
feat(zeroshot-rust): add native export and configuration inspection CLI commands


tomdps:
## Parent

- #687 — feat(zeroshot-rust): add the native CLI over the local daemon

## Execution

**AFK.** Start only after every issue in **Blocked by** is merged, required CI is green for the resulting `main` SHA, and the worktree is created from a freshly fetched `origin/main`.

## Outcome

Add native product export and effective configuration/profile inspection commands over the daemon/product configuration boundaries.

## Authoritative contracts/current state

#688 owns the bounded secret-free export service, #689 owns typed native configuration, #767 owns safe product error projection, and #765 establishes management CLI conventions.

## Owned scope

- Implement `export` as a bounded chunked daemon/product-control download with digest verification and atomic caller-selected file publication; never force 64 MiB through one WebSocket frame.
- Implement read-only configuration/profile inspection using redacted typed projections; secret values, handles, paths outside approved public roots, and provider raw metadata are unrepresentable.
- Use stable text/JSON renderers and product-local command/error types; no direct ledger/config-file mutation.
- Prove interruption/resume or restart behavior exactly as #688 exposes it and fail closed on digest/length mismatch.

## Acceptance criteria

- [ ] Exported bytes match the one #688 canonical bundle and never include forbidden raw records.
- [ ] Interrupted writes leave no partial destination and cannot replace an existing file without the declared policy.
- [ ] Config inspection shows precedence/source identity without secret material.
- [ ] Node package/CLI/configuration and `~/.zeroshot` remain untouched.

## Verification

```bash
cargo test -p zeroshot-rust --test cli_export
cargo test -p zeroshot-rust --test cli_config
cargo test -p zeroshot-rust --test architecture
npm run rust:check
npm run protocol:check
npm run typecheck
npm run lint
npm run test:unit
opcore check --changed
```

Every named focused target is an issue-owned deliverable and must execute non-zero tests before the common gates.

## Non-goals

- Run/management/lifecycle command implementation.
- Export bundle construction or daemon hosting.
- Configuration writes, migration, Node fallback, or package publication.

## Blocked by

- #765 — feat(zeroshot-rust): add native daemon and cluster management CLI commands
- #688 — feat(zeroshot-rust): persist execution evidence and add deterministic export
- #689 — feat(zeroshot-rust): add native config, binding references, and credential resolution
- #767 — feat(zeroshot-rust): project safe errors across product command and daemon boundaries

## Unblocks

[...]

Issue #748, opened 2026-07-23
feat(cluster-protocol): publish the TypeScript cluster client with full v1 parity


tomdps:
## Parent

- #651 — feat(cluster-protocol): ship WebSocket binding and TypeScript client parity

## Execution

**AFK.** Start only after every issue in **Blocked by** is merged, required CI is green for the resulting `main` SHA, and the worktree is created from a freshly fetched `origin/main`.

## Outcome

Publish `@the-open-engine/zeroshot/cluster` as the generated TypeScript full-v1 client after the complete Rust transport and optional subscription contracts exist.

## Authoritative contracts/current state

#651 owns the Rust WebSocket binding. #653 and #749 own cursorless logs and read-only attach. Rust/OpenRPC artifacts remain the only wire authority; the current package has no built cluster subpath and Node 18 cannot assume a global WebSocket.

## Owned scope

- Generate TypeScript wire types from authoritative protocol artifacts; hand-maintained duplicate DTOs are forbidden.
- Implement all unary methods and typed async subscriptions over WebSocket, including abort, iterator-return cleanup, reconnect for durable watch, and no replay for logs/attach.
- Publish a real CommonJS/ESM/type-resolvable `cluster` subpath on Node >=18 with an installed WebSocket runtime and injectable browser-compatible factory.
- Add deterministic cross-language goldens and packed-tarball import tests while preserving the existing default entrypoint and deep imports.

## Acceptance criteria

- [ ] Every Rust client method/event/error has a TypeScript parity case.
- [ ] Watch reconnect performs coherent `get` + `watch(fromCursor)` and deduplicates `(runId,cursor)`.
- [ ] Abort and iterator return cancel exactly once; logs/attach never claim a recoverable gap.
- [ ] Packed CJS, ESM, and TypeScript consumers import the subpath without changing current CLI/package behavior.

## Verification

```bash
npm run protocol:check
npm run typecheck
npm run lint
npm run test:cluster-client
npm run test:cluster-package
npm test
npm pack --dry-run
opcore check --changed
```

Every named focused target is an issue-owned deliverable and must execute non-zero tests before the common gates.

## Non-goals

- Rust dispatcher/server/client semantics.
- TypeScript in-process transport.
- Hosted authentication, provisioning, or native daemon behavior.

## Blocked by

- #651 — feat(cluster-protocol): ship WebSocket binding and TypeScript client parity
- #749 — feat(cluster-protocol): add capability-gated read-only agent attach

## Unblocks

[...]

Issue #691, opened 2026-07-16
test(zeroshot-rust): certify standalone product parity and namespace isolation


mkceichelbeck:
## Parent

- #665 — [EPIC] Zeroshot 2: separate native Rust engine

## Execution

**AFK.** Start only after every issue in **Blocked by** is merged, required CI is green for the resulting `main` SHA, and the worktree is created from a freshly fetched `origin/main`.

## Outcome

Certify the separately usable native product’s functional parity, namespace isolation, packaging, and release readiness after backend, built-ins, and every CLI slice are complete.

## Authoritative contracts/current state

#693 certifies embeddable full-v1, #700 certifies graph-visible proof/delivery, and #766 closes run/management/export/config CLI parity. This is a final audit/gate—not an implementation owner.

## Owned scope

- Run the pinned Node parity matrix for inputs, workflow routing, providers/models/reasoning, workspaces, PR/check/ship, lifecycle/monitoring/export/config, timeouts, and issue/source provider separation using native APIs only.
- Prove binary/library/config/data/cache/runtime/env/process/port/package identities are separate; Rust never reads/invokes/imports Node and both can run concurrently.
- Run clean-install/package/binary help/smoke tests and one backend invocation per supported provider family using fakes; record exact merged dependency SHAs and commands in a deterministic final manifest.
- Audit every acceptance item from #665; failures return to the owning issue and this PR may change only tests/manifests/package/release wiring.

## Acceptance criteria

- [ ] Every required parity row has observable evidence and no skipped/placeholder/mocked product path beyond declared fake external services.
- [ ] `zeroshot-rust --help` and all command smoke tests produce documented output on a clean machine fixture.
- [ ] Node package/CLI/settings/defaults remain byte-for-byte behaviorally unchanged.
- [ ] All repository gates pass at one exact dev SHA and both epics can close without unmerged release-critical work.

## Verification

```bash
cargo test -p zeroshot-rust --test standalone_parity
cargo test -p zeroshot-rust --test namespace_isolation
cargo test -p zeroshot-rust --test package_install
cargo test -p zeroshot-rust --test release_manifest
npm run rust:check
npm run protocol:check
npm run check
npm test
npm pack --dry-run
opcore check --changed
```

Every named focused target is an issue-owned deliverable and must execute non-zero tests before the common gates.

## Non-goals

- Implementing or repairing backend/driver/provider/daemon/CLI semantics.
- Node migration/default switch/deprecation/removal.
[...]

Issue #689, opened 2026-07-16
feat(zeroshot-rust): parse isolated native configuration and named profiles


mkceichelbeck:
## Parent

- #665 — [EPIC] Zeroshot 2: separate native Rust engine

## Execution

**AFK.** Start only after every issue in **Blocked by** is merged, required CI is green for the resulting `main` SHA, and the worktree is created from a freshly fetched `origin/main`.

## Outcome

Parse isolated native configuration directories/environment/CLI overrides into bounded typed settings and named profile definitions.

## Authoritative contracts/current state

#670 owns the provider catalog. Role contracts, admission manifests/bindings, and credential acquisition are split to #768/#769/#770 so this issue remains one configuration-edge PR.

## Owned scope

- Use separate OS-native Zeroshot Rust config/data/cache/runtime namespaces and only `ZEROSHOT_RUST_*`; never discover/import/open Node settings, `~/.zeroshot`, npm config, or legacy aliases.
- Implement precedence: explicit CLI option > ZEROSHOT_RUST_* env > selected native named profile > Rust catalog/product default; malformed present values fail with bounded corrective error.
- Define bounded versioned settings/profile schema for provider/source/issue/workspace/gateway/daemon/policy references using logical credential requirement names only.
- Canonicalize/validate profile inheritance or prohibit it explicitly, detect duplicates/cycles/unknown fields, and produce deterministic secret-free profile digests.

## Acceptance criteria

- [ ] Precedence, default, malformed, duplicate/unknown/cycle, path namespace, and cross-profile independence tests are exhaustive.
- [ ] Node settings/environment/files are never read even when present and both products coexist.
- [ ] Config/debug/error serialization contains no secret values or resolved capabilities.
- [ ] No role prompts, WorkerRegistry/manifest, credential lease, driver, daemon, or CLI command behavior is added.

## Verification

```bash
cargo test -p zeroshot-rust --test native_config
cargo test -p zeroshot-rust --test native_profiles
cargo test -p zeroshot-rust --test namespace_isolation
cargo test -p zeroshot-rust --test architecture
npm run rust:check
npm run protocol:check
npm run typecheck
npm run lint
npm run test:unit
opcore check --changed
```

Every named focused target is an issue-owned deliverable and must execute non-zero tests before the common gates.

## Non-goals

- Role contract pack.
- Admission manifest/worker binding compilation.
- Credential acquisition or provider/source/issue execution.

## Blocked by

[...]

PR #834, opened 2026-07-29
docs(v2): add parity matrix, launch order, behaviour-spec schema, and #748 design

tomdps:
Planning artifacts for epics #643 and #665, produced 2026-07-29 against Node baseline `50c5b00`. Docs only — no code, no behaviour change.

## What lands

| File | Purpose |
| --- | --- |
| `docs/v2/parity-matrix.md` | 281 capability rows (183 control + 98 engine), each classified `implemented`/`replaced`/`excluded`/`absent-in-node` and mapped to an owning #665 issue. Input artifact for #691. |
| `docs/v2/m1-launch-order.md` | Wave structure and launch rule for milestone `M1 — Certified embeddable backend`. |
| `docs/v2/behavior-spec-schema.md` | Scenario schema for the dual-binary suite run against both `zeroshot` and `zeroshot-rust`. |
| `docs/v2/748-typescript-client-design.md` | Ownership design gating #748. |

## Why the parity matrix matters

The Node test suite cannot serve as the V2 regression net: 159 of 214 test files `require()` a `src`/`lib`/`cli` module directly, and only `tests/e2e` (10 files) drives the real binary. #665 additionally forbids the Rust product from importing Node, so no shared harness is possible. This matrix plus the protocol conformance vectors are the actual net.

It is **provisional, not pinned** — Node keeps shipping and freezes later, so §9 carries a delta log that every Node feature PR must append to until the freeze.

## Findings worth reviewing

- **§7 records 71 unowned capabilities**, tiered by risk. Deliberately not filed as individual issues; this document is the record. Tier 1 includes that the V2 product currently has **no configuration write path at all** — #766's non-goals forbid writes and #689 is read-only, so a user cannot set a default provider.
- **§3.3** separates `excluded by decision` from `excluded by consequence`. The second kind is the dangerous one: nobody decided, it fell out of another constraint, and no owner recorded the loss.
- **§3.4** lists two contradictions to resolve before rows can be certified, plus two Node behaviours recorded as do-not-reproduce.

## CLAUDE.md corrections

Four claims were false and are fixed: the `enforce-main-pr-source` CI job does not exist in `.github/workflows/`; `dev` was deleted in `e94b6c2` so PRs target `main`; `.husky/pre-commit` does not validate that a test file exists; and conductor classification does not live in `src/conductor-bootstrap.js`. The "Where to Look" table re-pads because one entry got longer — prettier reflow, not a content change.

https://claude.ai/code/session_0133YEUjbWi2sNj85gmqt4hm


mkceichelbeck:

PR #857, opened 2026-07-30
feat(ledger): sign validator verdicts with offline-verifiable receipts

### LINKED ISSUE #464: Feature: Cryptographic Receipts for Validator Verdicts
## Problem

Zeroshot's blind validation model is brilliant — validators never see the worker's context, so their verdicts are genuinely independent. But the verdicts themselves are stored in SQLite as unsigned records. If the ledger is tampered with (or the machine is compromised), there's no way to prove what a validator actually decided.

For enterprise deployments where Zeroshot orchestrates production changes, teams need to answer: *"Can you prove this change was independently validated before it shipped?"*

## Proposal

Integrate [protect-mcp](https://www.npmjs.com/package/protect-mcp) to sign validator verdicts as Ed25519 receipts. Each validation cycle produces a cryptographic proof chain:

```
Plan receipt → Implementation receipt → Validator verdict receipt (signed)
                                        ├── tool: "validate"
                                        ├── decision: "pass" | "fail"
                                        ├── validator_id: (blinded)
                                        ├── issue_ref: "GH-123"
                                        └── signature: Ed25519
```

### What this enables

1. **Tamper-evident validation** — Each verdict is Ed25519-signed. If someone edits the SQLite ledger after the fact, the signatures break.
2. **Offline verification** — Anyone can verify the receipt chain without access to the Zeroshot instance: `npx @veritasacta/verify check ./receipts/`
3. **Issuer-blind verification** — Verifiers don't need to trust (or even know) who the issuer was. The math is the proof.
4. **Compliance evidence** — For teams using Zeroshot in regulated environments, signed receipts satisfy SOC 2 CC7.2 (system operation monitoring) and EU AI Act Article 12 (record-keeping).

### Integration surface

The natural hook points in Zeroshot's conductor loop:

| Event | Receipt type | Content |
|-------|-------------|---------|
| Plan generated | `plan_receipt` | Plan hash, planner agent ID |
| Implementation complete | `implementation_receipt` | Diff hash, worker agent ID |
| Validator verdict | `validation_receipt` | Pass/fail, validator ID (blinded), failure reasons |
| Cycle complete | `cycle_receipt` | Links all above, final status |

### Minimal integration

protect-mcp runs as a Claude Code HTTP hook server. If Zeroshot's agents run via Claude Code, the integration is two lines in `.claude/settings.json`:

```bash
npx protect-mcp init-hooks  # configures Claude Code to POST to protect-mcp
npx protect-mcp serve --enforce  # starts the receipt server
```

For non-Claude-Code backends (Codex, Gemini), protect-mcp also works as an MCP gateway wrapper or can be called directly via its Node.js API.

## Context

- [npm: protect-mcp](https://www.npmjs.com/package/protect-mcp) — MIT licensed, 0 config required
- [IETF Internet-Draft: draft-farley-acta-signed-receipts](https://datatracker.ietf.org/doc/draft-farley-acta-signed-receipts/) — The receipt format is an open standard
- [Merged into Microsoft Agent Governance Toolkit](https://github.com/AzureAI-Foundry/agent-governance-toolkit/pull/667)
- Uses [Cedar](https://www.cedarpolicy.com/) (AWS's authorization engine) for policy enforcement

Happy to build a proof-of-concept if there's interest. Zeroshot's blind validation + cryptographic receipts would be a strong combination — verifiable independent validation is a hard thing to find in the agent ecosystem.

tomjwxf:
Closes #464.

Thanks for the triage on 29 July. It named three things an implementation would need, and this is built to those rather than to my original proposal in the issue, which leaned on Claude Code hooks and on a package I maintain. Neither is a dependency here.

## The problem

Validator verdicts are the one ledger record a third party may later be asked to rely on: was this change independently validated before it shipped? They are stored unsigned in `messages`, so anyone who can write the database can change the answer afterwards and nothing in the ledger shows it.

## What this does

Signs each verdict when it is recorded and chains it to its predecessor, so editing, deleting or reordering is detectable by someone holding only the exported receipts and a published keyring.

**Provider independent.** Signing happens where verdicts are recorded, so it behaves identically whether the backend is Claude Code, Codex, Gemini or a local model. No vendor hook mechanism, no external service.

**Explicit key trust and rotation.** Keys live in a keyring with validity windows and revocation. A verifier resolves the key a receipt names and checks the receipt falls inside that key's window. Two decisions worth arguing with:

* Key ids derive from the public key, so a `kid` cannot be reassigned and always checks against the key it names.
* Revocation is deliberately **not** retroactive. A verdict signed while the key was trusted stays valid, because the alternative silently voids honest history every time an operator rotates. `revoked_at` exists for the compromise case, where you do want everything from that moment onward to fail.

**Offline verification.** Receipts plus a keyring. No cluster, no network, no service.

**No new dependencies.** Ed25519 via `node:crypto`.

## Purely additive

No existing file is modified. `git status` on this branch shows six new files and nothing else.

The store attaches to the ledger's existing `topic:VALIDATION_RESULT` event, so a deployment that never constructs one behaves exactly as today: no table, no signing, no cost. A test asserts that the receipts table is not created unless a store exists.

## The case this exists for

A deleted verdict leaves every remaining signature genuine, so signature checking alone reports success:

```
Cluster        demo
Receipts       2
Signatures     2 valid, 0 invalid
Sequence       ALTERED

Failures
  [1] v3
      sequence:  expected prev_receipt_hash 092971af... but found 9bf263de...

NOT VERIFIED
```

Signature validity and sequence integrity are reported separately on purpose. They are different failures calling for different responses, and collapsing them into one boolean would hide the distinction that makes the chain worth having. One break is reported once rather than cascading into every later receipt.

## Refusing rather than guessing

The decision is read from `content.data.approved`, the same field `applyValidationResult` already treats as the verdict, and it must be a real boolean. `"yes"`, `1`, `"true"` and a missing field all cause a refusal.

An unsigned verdict is a visible gap, and `findUnsignedVerdicts()` surfaces it. A confidently signed wrong one is worse, because the signature would verify perfectly over the wrong answer forever.

Refusals go to an `onError` callback rather than throwing, since this runs inside the ledger's `emit` and a throw would surface as the failure of an unrelated append.

## Known limitation, stated rather than hidden

Event attachment writes the receipt immediately after the verdict row, not inside the same transaction. A process killed in that gap leaves a verdict with no receipt. That is why `findUnsignedVerdicts()` exists. `recordVerdict(message)` is available for callers that want the write under their own control.

## What a verified receipt does not establish

Returned in every verification report rather than left to a reader to infer:

* Not that a verdict was correct.
* Not that the validator was independent or competent. Blind validation is a property of how Zeroshot runs validators, not something a signature confers.
* Not that verdicts signed by a compromised key were honest.
* Not that no verdict is missing from the front of the sequence. A chain proves nothing was removed from the middle or end of what you hold. It cannot prove you were given the beginning.

## Testing

39 new tests, most of them adversarial: payload tampering, signing with a second key while claiming a trusted `kid`, signing outside a validity window, revocation ordering in both directions, deletion, reordering, cross domain signature replay, and truthy stand ins for `approved`.

Verified locally against Node 22: 39 passing, plus 68 passing alongside the existing ledger suite. eslint clean, prettier clean, conventional commit per commitlint.

## Disclosure

I opened #464, I maintain protect-mcp, and I author the Acta signed receipts Internet-Draft whose format it follows. This implementation depends on none of them, and the source and tests reference none of them. protect-mcp appears once in `docs/signed-validator-verdicts.md`, as the fourth of four key custody options after file or environment, cloud KMS, and HSM, with that relationship stated in the doc. If you would rather it were not mentioned at all, deleting one bullet and one paragraph removes it with no effect on the implementation.

Happy to split this, rename anything, or move the store under `src/agent/` if that fits the layout better.


tomdps (changes requested):
Thanks for reworking this around the earlier feedback and for being explicit about the trust limits. The implementation and adversarial tests are thoughtful, but I can't approve this contract yet.

Three things need to change before this can merge:

1. The offline verifier needs a trusted completeness anchor (for example a signed manifest/checkpoint containing the expected count and terminal receipt hash). Internal predecessor links cannot detect suffix deletion or an empty bundle, so the current result can say VERIFIED after the newest receipts are removed.
2. Chain order must come from durable append/signing order, not message timestamps. An older timestamp recorded later currently makes honest evidence look altered.
3. This needs to be integrated at the authoritative verdict persistence boundary. As submitted, the store is opt-in and no product path constructs it, while event attachment leaves a non-atomic verdict/receipt gap. That is useful library code, but not a delivered Zeroshot guarantee.

I'm leaving #464 open because the capability is still worth pursuing. Please resolve the trust/completeness contract there first if the atomic persistence change makes this substantially larger; I would rather review the right design than ask you to patch around the current shape.

tomjwxf:
All three findings are correct and I am not going to argue any of them.

I reproduced the two concrete ones against this branch before replying:

```
empty bundle       -> valid=true  receipts_checked=0
newest verdict cut -> valid=true  chain_intact=true  signatures_invalid=0
```

The suffix case is the one that matters. I led this PR with a middle deletion being caught and documented that a chain cannot prove you were given the beginning, but I never considered the end, which is the easier attack and the likelier one.

Your third point is the one I had backwards. I optimised for touching no existing files and presented that as a strength. Zero files modified also means no product path uses it, which makes it library code rather than a guarantee Zeroshot delivers.

Marking this draft rather than patching around the current shape. Design proposal in #464 covering the completeness anchor, durable ordering, and the persistence boundary, with the anchor question first since that is the part your review left open and the part that decides whether the rest is worth having.

mkceichelbeck:

PR #863, opened 2026-07-31
docs: make the README match the engine

EivMeyer:
Read the README against `src/config-router.js`, `cluster-templates/base-templates/`, `src/agent-cli-provider/provider-registry.ts` and `src/issue-providers/`. Two claims turned out not to be true, four contradicted either the engine or the README itself, and a fair amount of what Zeroshot does wasn't mentioned at all.

## Claims that weren't true

**"an independent verifier that never saw how it was made."** Validators receive `PLAN_READY` and `IMPLEMENTATION_READY`, so they see the plan and the executor's own summary of the work. What they never see is it working: `WORKER_PROGRESS` is absent from every validator's `contextStrategy` and sessions aren't shared. The page now claims that instead, which is both true and harder for anyone else to claim.

**"a reproducible failure."** Fits `validator-tester` failing a suite. It doesn't fit `validator-code` rejecting a silently swallowed error, or `validator-security` hitting its instant-reject list; those return free-text `errors[]` and there's no command to re-run. Replaced with specificity, which the comparison table was already reaching for.

## Contradictions

- "How it works" put the executor in an isolated workspace, while the Isolation section says the default is `None`.
- TRIVIAL routes to `single-worker` and `getValidatorCount()` returns `0`, so the headline executor-verifier split doesn't apply on that path unless `--pr` or `--ship` adds a validator back. Now stated directly under the comparison table.
- Non-TRIVIAL DEBUG routes to `debug-workflow` (investigator, fixer, tester, completion-detector), which has no validator role. The old text described the TASK/INQUIRY lane as though it were universal, and "fix a login bug" is one of the README's own good-fit examples.
- "stores no API keys and manages no auth" is false for Linear and for `gateway`, both of which keep a key in settings. Scoped to provider CLIs, where it holds.

## What was missing

Four of the eight providers, `gateway` among them. It speaks the OpenAI or Anthropic wire protocol against any `baseUrl`, so Zeroshot can drive a self-hosted model or a proxy with no coding-agent CLI installed. Nothing was hiding these: `VALID_PROVIDERS` is spread straight from `registry.providerIds` with no filter, so `zeroshot providers` has always listed all eight and `docs/providers.md` documents them. Only the README was behind.

Linear, the fifth issue backend, with the `KEY-NUMBER` ambiguity against Jira and how it resolves.

A routing table, so the validator count per classification is on the page rather than inferable from `config-router.js`.

`--config` and the `zeroshot config` commands, which appeared zero times:

```
grep -cE '\-\-config|zeroshot config' README.md  →  0
```

That gap mattered most. Agent ids, roles and topic names are free strings, triggers carry arbitrary JS predicates, cycles are legal so long as a ring of three or more has escape logic, and sub-clusters nest five deep. The README read as a fixed pipeline when the runtime underneath is a general graph engine, so there's now a second paragraph at the top saying so. It leads on determinism, because that's the part nothing else claims: `logic-engine.js` evaluates trigger predicates in a `vm` sandbox, so once the conductor has classified, no model is anywhere in the scheduling path.

Two examples also labelled a bare issue number as GitHub. Resolution actually runs force flags, then the git remote, then `defaultIssueSource`, and only falls back to GitHub when there's no git context at all, so `zeroshot run 123` in a GitLab checkout opens a GitLab issue.

## Rewrites

The lede was singular where the engine is plural, and never mentioned the conductor, while the graph immediately beneath it opens with a classification matrix and two conductors. "How it works" had grown to six long sentences doing seven jobs, two of which repeated the lede word for word.

The last commit is tidying after those rewrites rather than new claims: two code fences had ended up back to back in Providers, "validator" was being used both for a review agent and for `zeroshot config validate`, the architecture `<details>` block still carried a compressed summary of routing that the new section now covers properly, and the same point about the shipped graph not being the only one was being made twice, two hundred lines apart.

## One question on the diff

`pi` pins `@0.80.3` and passes `--ignore-scripts`. It's in the provider table because `docs/providers.md` already advertises it that way, but say the word and I'll pull the row.

No behaviour changes. Beyond the README: `CLAUDE.md`'s validator counts, which disagreed with both the router and the shipped conductor prompt, and one `description` string in `conductor-bootstrap.json` that said level1/level2 where the agents are level2/level3. Prettier clean.


EivMeyer:
Superseded by #1014 and #1015.

This branch was cut on 2026-07-31 and #945 rewrote the README under it on 2026-08-05, so most of what's here is either already fixed on main or now wrong. The provider table and the Linear/issue-source expansion landed with #945. Node ≥ 18 is stale (main says 22), and "by default it edits your files in place" is backwards now that guided setup defaults fresh repos to worktree isolation. The lede and "How it works" rewrites would revert #945's voice.

What still held, split out and re-verified against main at 7be7eea:

- #1014: the validator counts in CLAUDE.md and AGENTS.md, plus the level1/level2 description in conductor-bootstrap.json
- #1015: the routing table, the `zeroshot config` section, and the "reproducible failure" claim

One correction to this PR's description while closing it: `debug-workflow` does have a validator-role agent, `tester`. The shape is different from executor–verifier, but the claim that it "has no validator role" was wrong.

PR #872, opened 2026-07-31
feat(docker): support OMP execution and credentials in isolation

### LINKED ISSUE #868: feat(omp): add Docker distribution and broker auth
Parent: #864

Depends on: #900

## Objective

Add Docker distribution and env/broker authentication.

## Current released state

## Scope

- Extend registry-derived Docker presets to support an env-only provider with zero automatic mounts. Keep full adapter credential inventory distinct from the exact automatic Docker env allowlist, and base missing-credential diagnostics on the effective container env/mount plan rather than host presence.
- The current cluster Dockerfile is an amd64 image in practice: its AWS, Terraform, kubectl, Helm, Infracost, TFLint, and tfsec layers hard-code x86-64 assets. Do not claim a native arm64 OMP image in this epic.
- Add optional registry-owned `docker.platform`; set OMP to `linux/amd64` and pass it to both image build and container run. Existing providers retain their current unset behavior. Fail before workspace/container effects when the engine cannot run that platform; a separate base-image multiarch issue must convert every hard-coded tool before OMP can advertise native arm64 Docker.
- The pre-effect probe requires a Linux Docker server. Native `amd64` server architecture satisfies the registry platform directly; any other architecture must have a functioning Buildx builder whose advertised platforms include `linux/amd64`. Missing Buildx/emulation fails with remediation before worktree or container creation; the subsequent image build's pinned `omp --version` step is the execution proof.
- Add the OMP Docker variant with exact `omp-linux-x64` from the release contract, verify SHA-256 `ac0285a571aa79c58d59482561a3871befe7333dba3a3bdc2e90682653ee33b2` before installation, and require runtime version `17.2.1`; never fetch `latest` or trust a tag URL alone.

- Run OMP as the existing non-root container user. Create its config/overlay roots owner-only inside the container; do not mount/copy the host OMP auth store.
- Keep Docker execution explicitly fresh-only: pass `--no-session`, retain `agentCanReuseSession`'s fail-closed isolation rule, do not persist an OMP continuation from an isolated task, and rebuild full context after every task/container recreation. Do not claim same-container or recreated-container resume.
- Add the exact environment/auth-broker allowlist above, missing-auth/custom-credential remediation, and secret redaction.
- Host-to-container live guidance remains `unsupported('isolation-enabled')`; queued guidance is consumed in the next fresh full-context safe point.
- Flip only `omp.capabilities.dockerIsolation` to true after Docker tests pass.
- Update `AGENTS.md` with the env/broker-only, sessionless OMP Docker boundary here.

## Primary files

- `src/agent-cli-provider/provider-registry.ts`
- `lib/docker-config.js`, Docker preset tests
- `docker/zeroshot-cluster/Dockerfile`
- `src/isolation-manager.js`, isolated task command/env paths
- `src/agent/provider-session.js`
- Docker auth/isolation/fresh-only recreation tests and `AGENTS.md`

## Acceptance

- The `linux/amd64` build verifies the exact digest and `omp --version` returns exactly 17.2.1; build and run receive the same registry-derived platform. An emulated arm64 host either proves that platform available or fails before effects—never silently installs the arm64 binary into the mixed-architecture base.
- API-key and auth-broker fixtures work; empty/malformed auth fails noninteractively with actionable remediation and no provider fallback.
- `~/.omp`, `agent.db`, WAL/SHM files, and host refresh tokens are not automatically mounted or copied.
- Custom env/path credentials require explicit passthrough/mount and are identified as such.
- Container config/overlay roots are owner-only, OMP runs non-root, cancellation/reaping remains bounded, and cleanup ownership survives daemon recovery.
- Isolated session reuse remains rejected before resume construction; same-container next task and container recreation both send full context and never reference a prior OMP session ID.
- Existing provider Docker presets remain unchanged.

## Verification

`npm run test:omp` plus focused Docker preset/image/preflight tests. Do not run a paid model request.

## Source contract

[v17.2.1 release assets](https://api.github.com/repos/can1357/oh-my-pi/releases/tags/v17.2.1), [environment variables](https://github.com/can1357/oh-my-pi/blob/v17.2.1/docs/environment-variables.md), [auth broker](https://github.com/can1357/oh-my-pi/blob/v17.2.1/docs/auth-broker-gateway.md)

## Delivery constraints

tomdps:
Closes #868


mkceichelbeck:

PR #879, opened 2026-08-01
fix: close daemon profile creation race

EivMeyer:
## Summary

- create Unix daemon profile directories with owner-only mode in the initial mkdir operation
- restore exact owner access after umask filtering without ever exposing group or other access
- add 32-contender regressions for simultaneous first-time creation and restrictive-umask creation
- keep concurrent listener startup fail-closed when authenticated liveness is temporarily indeterminate

## Why

Two starters could both observe a missing profile directory. The winner created it with default permissions and only tightened it afterward, so the other starter could transiently reject the directory as insecure. This was surfaced by the concurrent daemon-listener test in PR #863.

The listener assertion also assumed every safe loser must report AlreadyRunning, although the documented liveness contract permits LivenessIndeterminate during the publication-to-accept-loop handoff. Both outcomes preserve the winning locator; discovery and security errors remain rejected.

## Validation

- cargo fmt --all -- --check
- cargo clippy -p zeroshot-rust --test daemon_discovery --test daemon_listener
- cargo test -p zeroshot-rust --test daemon_discovery (10 passed)
- cargo test -p zeroshot-rust --test daemon_listener (19 passed)
- concurrent profile-creation regression: 1,000 consecutive passes
- restrictive-umask concurrent-creation regression: 500 consecutive passes
- concurrent listener-start regression: 1,000 consecutive passes

### DIFF
--- a/zeroshot-rust/src/daemon_discovery.rs
+++ b/zeroshot-rust/src/daemon_discovery.rs
@@ -261,7 +261,7 @@ fn secure_directory_handle_shape(is_directory: bool, is_reparse_point: bool, lin
 mod platform {
     use std::fs::{self, File, OpenOptions};
     use std::io;
-    use std::os::unix::fs::{MetadataExt, OpenOptionsExt, PermissionsExt};
+    use std::os::unix::fs::{DirBuilderExt, MetadataExt, OpenOptionsExt};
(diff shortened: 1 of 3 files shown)


mkceichelbeck:

PR #908, opened 2026-08-02
feat: feat(omp): persist and verify resumable sessions

### LINKED ISSUE #866: feat(omp): persist and verify resumable sessions
Parent: #864

Depends on: #900

## Objective

Add durable OMP session ownership and verified resume.

## Current released state

## Scope

- Replace `--no-session` with `--session-dir <verified-partition>` for fresh supported host/worktree and standalone task flows. Verified continuation adds `--resume <verified-absolute-jsonl>`, never an ID search. Do not relocate `PI_CODING_AGENT_DIR`; Docker remains sessionless.
- Allocate a random, secret-free UUID partition at the trusted initial spawn boundary under the owning cluster `storageDir` or standalone `TASKS_DIR`. Persist it with cluster/agent/canonical-workspace or standalone-lineage bindings and pass it through `agent-task-executor`, `task-lib/runner.js`, detached watcher config, and `task-lib/commands/resume.js`; manual resume reuses it. Never derive ownership in the watcher from prompt text or cwd alone.
- Implement the two-phase lazy-file contract above: existing sessions and the sibling artifact tree are verified before spawn and before prompt; fresh sessions are path-checked at `ready` and descriptor/header/tree-verified after terminal materialization. Put these non-configurable bounds beside the verifier as `OMP_SESSION_LIMITS`: `{maxSessionBytes:268435456,maxSessionRecords:1000000,maxArtifactEntries:4096,maxArtifactDepth:16,maxRelativePathBytes:4096,maxArtifactFileBytes:268435456,maxArtifactAggregateBytes:536870912,maxBlobReferences:4096,maxReferencedBlobBytes:67108864}`. Stream JSONL, manifest files, and referenced CAS blobs; reject before allocation when a declared/observed bound is exceeded. Accept only owner-held directories and regular single-link files, reject symlinks/sockets/devices/hard links, hash a length-prefixed raw-path/type/size/content-digest manifest, require canonical `blob:sha256:<64-lower-hex>` references whose bytes match the digest, and treat a missing blob as an invalid continuation.
- Extend the existing `providerSession` snapshot with exact optional field `ompSession`. For provider `omp`, `normalizeProviderSession` requires `{schemaVersion:1,partitionId,sessionFileName,sessionFileIdentity:{device,inode},artifactManifestDigest,executionFingerprint,selectedProvider,selectedModel}` in addition to the existing provider/sessionId/agentId/taskId/generation/cwd/worktreePath/contextSequence/guidanceSequence/promptIdentity tuple; every digest is `sha256:<64-lower-hex>` and every device/inode is a canonical unsigned decimal string. For every other provider `ompSession` is absent. Do not persist storage-root or partition paths in the agent snapshot.
- Bump `task-lib/store.js` to schema version 5 and add one nullable `omp_session_ownership TEXT` column mapped to `task.ompSessionOwnership`. Encode only the following closed JSON object:

```json
{
  "schemaVersion": 1,
  "state": "provisional|committed|cleanup-required",
  "partitionId": "<uuid>",
  "storageRoot": "<canonical-absolute-path>",
  "partitionPath": "<canonical-direct-child-path>",
  "ownerUid": "<canonical-decimal>",
  "storageRootIdentity": { "device": "<decimal>", "inode": "<decimal>" },
  "partitionIdentity": { "device": "<decimal>", "inode": "<decimal>" },
  "canonicalWorkspace": "<canonical-absolute-path>",
  "owner": {
    "kind": "cluster-agent|standalone",
    "clusterId": "<string-or-null>",
    "agentId": "<string-or-null>",
    "taskId": "<string>"
  },
  "session": {
    "sessionId": "<string>",
    "fileName": "<direct-child-basename.jsonl>",
    "fileIdentity": { "device": "<decimal>", "inode": "<decimal>" },
    "artifactManifestDigest": "sha256:<hex>",
    "executionFingerprint": "sha256:<hex>",
    "selectedProvider": "<string>",
    "selectedModel": "<string>"
  }
}
```

`partitionIdentity` and `session` are `null` only while `state` is `provisional` or when an uncertain cleanup-required record never observed them; `committed` requires every field. `cluster-agent` requires non-null cluster/agent IDs; `standalone` requires both null. Validate and canonicalize this object on every read/write. Existing rows migrate to SQL `NULL` without fabricated values: non-OMP resume stays unchanged, while an OMP row or snapshot lacking its complete committed tuple fails closed to fresh context.
- Add owner-fenced task-store transitions: persist provisional before directory creation, CAS a committed lineage to the resumed task before prompt write, commit only after standalone output validation or cluster logical/schema/hook success, and mark cleanup-required on every failed, cancelled, or uncertain boundary. Recovery derives the decision from the exact durable task/lifecycle boundary; it never guesses from file presence.
- Wire the verified whole JSONL partition into cluster `clear`, standalone task `clean`, and global `purge`; committed sessions remain available for resume until that boundary, unsafe paths preserve the owner record with an actionable warning, and shared OMP CAS blobs are never deleted.
- Keep cluster guidance on its existing durable safe-point path and standalone live input on its existing unsupported behavior until Subissue 5. Session persistence must not silently imply acknowledged live steering.
- Update `AGENTS.md` with OMP session provenance, ownership, verification, and cleanup conventions here.

## Primary files

- OMP RPC session driver and pipe watcher
- `src/agent/provider-session.js`, `src/agent/agent-lifecycle.js`, and `task-lib/provider-session-capture.js`
- task record/storage-root/ownership-transition/spawn/resume codecs in `src/agent/agent-task-executor.js`, `task-lib/config.js`, `task-lib/store.js`, `task-lib/runner.js`, and `task-lib/commands/resume.js`
- cluster `clear`, standalone task `clean`, and global `purge` state-cleanup paths
- lifecycle/resume/ownership/cleanup tests and `AGENTS.md`

## Acceptance

- Fresh cluster-agent and standalone turns capture exactly one `{sessionId,sessionFileName,sessionFileIdentity,partitionId,artifactManifestDigest,executionFingerprint,selectedProvider,selectedModel}` value out of band; resumed turns require the exact persisted value. Manual standalone resume and detached recovery use the same partition under their injected storage root.
- Existing-but-wrong recorded cwd, moved/deleted workspace, symlink/non-regular/multiply-linked/substituted file, changed ID/path/inode, conflicting IDs, incomplete prior lifecycle, artifact-tree type/path/content/manifest mismatch or bounds, Zeroshot selector/thinking/overlay/version drift, selected concrete-model drift, worktree drift, or missing partition identity/partition/referenced blob never reaches a resume prompt. Zeroshot retires that owner and rebuilds in a new partition where its existing policy allows. Tests cover custom cluster `storageDir` and standalone `TASKS_DIR`; cleanup leaves the shared OMP blob root unchanged.
- Events-before-ack and nonterminal continuation sequences do not commit early.
- Ownership crash vectors cover row-before-directory, fresh materialization-before-capture, resume-transfer-before-prompt, provider-complete-before-hook, commit-before-agent-snapshot, failed schema/hook, cancellation, and cleanup failure. Each resolves to exactly one committed owner or retryable cleanup-required state; no unowned partition becomes resumable.
- Advertise `sessionResume:true` only after host, worktree, detached cluster-agent, and standalone manual-resume tests pass; document Docker as fresh-only.

## Verification

`npm run test:omp`

`npx mocha tests/unit/provider-session-capture.test.js tests/unit/provider-session-restore.test.js tests/unit/provider-session-reuse.test.js --timeout 30000`

## Source contract

[OMP session manager](https://github.com/can1357/oh-my-pi/blob/v17.2.1/packages/coding-agent/src/session/session-manager.ts), [RPC v2](https://github.com/can1357/oh-my-pi/blob/v17.2.1/docs/rpc.md)

## Delivery constraints

tomdps:
Closes #866


tomdps:
Blocking source-contract review; this PR is not mergeable yet.

1. `src/omp-session-verifier.js` implements an invented `<storageRoot>/omp-sessions/.blobs` store and treats pointer-only artifact files as CAS refs. OMP v17.2.1 actually writes nested `blob:sha256:<hex>` refs inside session JSONL and resolves them from `@oh-my-pi/pi-utils::getBlobsDir()` (`~/.omp/agent/blobs`, subject to OMP's config-root semantics). Parse/stream the JSONL records, collect canonical nested refs, and verify the referenced canonical blobs at the actual shared OMP root. Cleanup must never delete that root. Official sources: `packages/coding-agent/src/session/session-loader.ts`, `blob-store.ts`, and `packages/utils/src/dirs.ts`.
2. Cluster sessions never become reusable. `buildCompletionResult()` creates `result.providerSession` while the ownership row is still `provisional`, so `providerSessionFromCompletedTask()` returns null; `runTaskAttempt()` then commits the row and writes that stale null snapshot. Re-read/rebuild only after a checked successful commit and before publishing `TASK_COMPLETED`.
3. Resume does not transfer ownership. `resolveOmpSessionPlan()` creates a second provisional owner for the same partition without atomically relinquishing the prior committed owner; after success there are two committed task rows. Implement the issue's owner-fenced CAS transfer before prompt write and cover resume-transfer crash/re-entry.
4. The resume descriptor omits `sessionId`, `partitionIdentity`, and `executionFingerprint`. The watcher neither compares the echoed full session ID/path to the exact persisted tuple nor checks version/thinking/overlay execution drift. It can finalize a different returned file because it verifies `basename(result.session.sessionFile)` instead of requiring the requested path/name. Pass and compare the complete committed tuple; conflicting/missing/full-vs-prefix identity must fail before hooks.
5. Cleanup ignores provisional rows and deletes by path after a check without validating persisted owner UID, storage-root identity, or partition identity; a substituted directory can be recursively deleted. `cluster clear` and global `purge` are not wired at all (the only `cli/index.js` change is the hidden resume option). Implement all three required cleanup surfaces, preserve unsafe owner rows with warnings, and close the check/use race flagged by CodeQL.
6. The ownership validator accepts non-UUID partition IDs, non-direct/non-`.jsonl` file names, relative/noncanonical paths, and partially populated noncommitted identity/session pairs despite the closed schema. Validate/canonicalize every read/write and enforce row owner task ID.
7. The current file-path verification uses lstat/open/stat pathname sequences and recursive name-based traversal, leaving the CodeQL races real. Verification must pin descriptors/identities and stream from those opened files; do not re-open mutable names after validation.

Add focused tests for each point plus every #866 acceptance crash vector. `npm run test:omp` and the three provider-session suites are necessary but not sufficient unless they exercise host, worktree, detached cluster, standalone manual resume, custom storage roots, blob refs, and all cleanup surfaces.

tomdps:
Repaired on `zeroshot/bronze-beacon-35` (`376eec8`). Every point below is reworked against the tagged v17.2.1 source, not patched around.

---

**1. CAS root and refs.** The invented `<storageRoot>/omp-sessions/.blobs` store and the pointer-only-artifact-file model are gone. Per `session/blob-store.ts` and `session/session-loader.ts`, OMP leaves nested `blob:sha256:<64-lower-hex>` refs *inside* the JSONL records and resolves them from `pi-utils::getBlobsDir()`. New `src/omp-blob-root.js` mirrors `dirs.ts`'s resolution exactly — `PI_CONFIG_DIR`, `PI_CODING_AGENT_DIR` (default-mode only), `OMP_PROFILE`/`PI_PROFILE`, and the linux/darwin XDG_DATA_HOME redirect keyed on the profile-specific path. `src/omp-session-verifier.js` now streams *and parses* the JSONL, collects canonical nested refs, and verifies the blobs at that real shared root. A blob may legitimately carry `nlink > 1` (`ensureDisplayPath` hardlinks the typed sidecar), so blobs are content-verified rather than link-count-verified; a missing, non-canonical, or digest-mismatched ref is an invalid continuation. Cleanup can never reach that root — `deleteOmpSessionPartition` refuses any path resolving inside it.

**2. Cluster sessions were never reusable.** Confirmed and fixed. `buildCompletionResult()` necessarily computes `providerSession` while the row is still `provisional`, so it was `null` for OMP *by construction*. `src/agent/agent-lifecycle.js#finalizeProviderSessionAfterCommit` now commits, **checks** the commit applied, re-reads the row, and rebuilds the snapshot before it reaches the agent or `TASK_COMPLETED`; a commit that does not apply retires the partition. Proven directly rather than inferred — the test asserts `preCommitSnapshot === null` *and* a fully populated rebuilt snapshot from the same run:

```
$ npx mocha tests/unit/omp-provider-session.test.js --timeout 30000
  commit-then-snapshot ordering (agent-lifecycle.js)
    ✔ rebuilds the snapshot after the commit rather than publishing the stale provisional one
    ✔ retires the partition and yields no snapshot when there is no evidence to commit
```

**3. Owner transfer.** `resolveOmpSessionPlan()` no longer creates a second owner. `transferOmpSessionOwnership` moves the prior committed lineage onto the resumed row and clears the prior row **in one transaction, both sides fenced on their exact persisted JSON**; the watcher runs it from the `ready` hook strictly before the prompt write. The resumed row stays `provisional` until its own success boundary, so a partition never has two committed owners nor none, and a half-finished continuation is never published as resumable. Cleanup consults `findCommittedOwnersForPartition`, so a resume that crashed before its transfer cannot delete the session its still-committed prior owner can legitimately resume.

**4. Descriptor completeness.** The descriptor now carries `expectedSessionId`, `expectedExecutionFingerprint` and `priorOwnerTaskId`, and is cross-checked field-by-field against that persisted row in `task-lib/runner.js#resolveOmpResumeExpectation` *before a task row exists*. The watcher compares the **full** session id and the **full** session file path (`resolve(evidence.sessionFile) !== expected.sessionFilePath` — never `basename`), partition and session-file inode identity, manifest digest, the transcript header's own id and cwd, and an `executionFingerprint` (`src/omp-execution-fingerprint.js`) binding the pinned OMP release, the config-overlay content digest, the requested `--model`/`--thinking`/`--approval-mode`, and the concrete provider/model/thinking level. The "same basename, different directory" case has its own test.

**5. Cleanup surfaces.** All three now exist and share `task-lib/omp-session-cleanup.js`: task `clean`, cluster clear (`cli/index.js#deleteClusterData`, which reclaims partitions under the cluster's own `storageDir`), and global `purge` (cluster clear → `clean --all`). `provisional` rows are cleaned too, so no orphans. Deletion validates the persisted owner uid and the storage-root and partition inode identities, then closes the CodeQL check/use race by **staging before deleting**: the partition is renamed inside its own parent to an unguessable `.zeroshot-deleting-<uuid>` name and only then re-pinned and removed, so after the atomic rename the object cannot be swapped by racing the original path. A substituted directory is parked and reported, never recursively deleted; unsafe paths keep the owner record with a warning.

**6. Closed schema.** UUID partition ids, canonical absolute paths, a **re-derived** `partitionPath` (a stored one is never trusted), direct-child `*.jsonl` file names, no partially populated identity/session pairs in any state, no unknown keys at any level, and `validateOwnedByTask` fencing every record to the row it was read from. Transitions are full-value CAS, and `updateTask` now only writes the column when the caller explicitly supplies it — otherwise an unrelated row update could clobber a CAS another process had just performed.

**7. Verification races.** No lstat→open→stat sequence and no re-open of a mutable name survives. Each path is opened once with `O_NOFOLLOW|O_NONBLOCK` (plus `O_DIRECTORY`), and every type/owner/link/size/identity check and every byte read comes from `fstat`/`read` on that same descriptor; directory listings re-pin and compare identity around `readdir`, so a substituted directory is a hard failure rather than a silent traversal. `O_NONBLOCK` also fixes a real hang this work surfaced: opening a FIFO read-only blocks until a writer appears, so a FIFO planted in a partition previously hung verification forever instead of failing it.

**Also fixed, not in the list:** `session_info_update` re-fires the driver's `ready` hook *after* the prompt, where re-running the structural manifest/inode comparison would have rejected a perfectly healthy turn whose transcript had legitimately grown. Post-prompt passes now re-check only what OMP reports (identity + fingerprint), which still catches a mid-turn session switch. Covered by a test that grows the transcript mid-turn and asserts both the healthy and the switched case.

---

### Tests

Six new suites plus rewritten verifier/ownership/watcher suites. The fake `omp --mode rpc` now takes real `--session-dir`/`--resume` argv and materializes a real transcript whose first record is OMP's session header (`<fileSafeTimestamp>_<sessionId>.jsonl` + sibling artifacts dir, per `session-manager.ts`), so the watcher tests exercise the actual contract.

| Suite | Covers |
|---|---|
| `tests/unit/omp-blob-root.test.js` (11) | `getBlobsDir()` parity: default, `PI_CONFIG_DIR`, `PI_CODING_AGENT_DIR`, profiles, XDG existence-gating and profile keying, invalid profile names, blob-root containment guard |
| `tests/unit/omp-session-verifier.test.js` (34) | header parsing, nested refs at the real shared root, missing/mismatched/non-canonical/symlinked blobs, sidecar hardlink accepted, symlink/hardlink/FIFO/substituted-inode rejection, manifest determinism and content/path/type drift, every `OMP_SESSION_LIMITS` bound (pinned to the exact issue values) |
| `tests/unit/omp-session-ownership-schema.test.js` (15) | every closed-schema rejection, owner-task fencing, byte-stable canonical round-trip |
| `tests/unit/omp-session-ownership.test.js` (20) | two-boundary commit contract, all 8 #866 crash vectors, atomic transfer + idempotent replay + cross-partition refusal |
| `tests/unit/omp-session-cleanup.test.js` (19) | all three surfaces, uid/storage-root/partition identity refusals, symlink and substitution refusal, staging leaves nothing behind, custom cluster `storageDir` + custom standalone `TASKS_DIR`, shared blob root byte-identical across delete |
| `tests/unit/omp-resume-plan.test.js` (18) | descriptor emission → strict parse → row cross-check; conflicting IDs, wrong/moved cwd, storage-root drift, incomplete prior lifecycle, manual standalone resume |
| `tests/unit/omp-provider-session.test.js` (8) | exact-optional `ompSession` field set, commit-then-snapshot ordering |
| `tests/omp-rpc-watcher.test.js` (29) | host / worktree / detached cluster-agent fresh commits, ownership transfer before prompt, and drift: echoed session id, same-basename different path, concrete-model, thinking level, selector/overlay/version fingerprint, artifact tree, inode substitution, symlink, partition identity, missing blob, unavailable transfer, mid-turn `session_info_update` |

### Evidence

```
$ npm run test:omp
ℹ pass 108   ℹ fail 0            # node --test tests/agent-cli-provider/omp-*.test.js
34 passing (8s)                  # omp-rpc-watcher + omp-isolation-capability

$ npx mocha tests/unit/provider-session-capture.test.js \
            tests/unit/provider-session-restore.test.js \
            tests/unit/provider-session-reuse.test.js --timeout 30000
19 passing (142ms)

$ npx mocha tests/unit/omp-*.test.js --timeout 30000
125 passing (4s)

$ npm test
EXIT=0                           # 2323 passing, full unit suite

$ npm run check:agent-cli-provider
ℹ pass 274   ℹ fail 0            # typecheck + lint + build + provider tests

$ npm run lint
✖ 284 problems (0 errors, 284 warnings)   # 0 errors; warnings are pre-existing repo-wide

$ npm run typecheck
(clean — tsc --noEmit && tsc --project tsconfig.cluster.json)
```

`AGENTS.md` and `docs/providers.md` are rewritten for the corrected CAS, transfer, verification, and cleanup semantics.

One caveat worth flagging: the full parallel `npm test` run produced one load-induced failure in `tests/agent-stuck-recovery.test.js` (a 30s spawn-timeout expectation firing when the box is saturated). It is unrelated to this change and passes in isolation (`14 passing`); `npm test` still exits 0.

tomdps:
Follow-up pushed (`3ccc2cf`), which also merges `main` — #909 landed while I was working, so the PR was `CONFLICTING` and GitHub was refusing to run `pull_request` checks at all. They're running now.

**Conflict resolution.** #909 flipped OMP's `dockerIsolation` to `true`; #866 flips `sessionResume` to `true`. Both stay — they're independent — and the registry comment plus `AGENTS.md` now say why rather than leaving the reader to infer it.

**That merge exposed something worth flagging.** Before #909, `dockerIsolation: false` meant OMP + Docker was rejected at preflight, so "Docker remains sessionless" was moot. It isn't now. A partition allocated inside a container could never be resumed, and its ownership row would be unreclaimable the moment the container is removed — so an isolated run now carries `ZEROSHOT_OMP_SESSIONLESS=1`, `resolveOmpSessionPlan` allocates nothing and persists no ownership row, and the adapter falls back to `--no-session`. `tests/omp-docker-fresh-only.test.js` is updated from main's "sessionless for every OMP task" assertions to the real post-#866 contract, and now also pins the *positive* side (an un-isolated agent with a complete verified snapshot does reuse), so a future change can't quietly re-disable resume everywhere and still pass that suite.

**A cleanup-fencing bug, found and fixed.** The version I described in my previous comment fenced cleanup on *committed* owners. That is wrong in precisely the state that matters. More than one row can name one partition — a resumed row is inserted before its transfer runs, and two competing resumes of one committed session leave three rows on it (prior owner + both candidates). Right after a successful transfer the **winner is `provisional` and no row is committed at all**, so a committed-only fence returned nothing and a retired losing competitor would have renamed and deleted the winner's live session. The loser's own record carries no `partitionIdentity` either, so the descriptor-pinned identity check wouldn't have caught it on the way through.

`findAuthoritativeOwnersForPartition` now fences on every `provisional`-or-`committed` claim other than the acting row's. `cleanup-required` is deliberately *not* authoritative — treating it as one would deadlock two retired rows against each other and strand the partition permanently. Regression test:

```
✔ third-owner interleaving: the retired loser is the only row a committed-only fence can see
✔ preserves the partition of a resume that WON the transfer against a losing competitor
```

**Deletion is also now properly fenced in time**, not just in state. It splits into `stageOmpSessionPartitionForDeletion` (validate + atomic rename) and `removeStagedOmpSessionPartition` (`rm -r`), with the owner check and the rename inside one `BEGIN IMMEDIATE` task-store transaction. Without that, a resume could insert its row or win its transfer in the gap between check and rename; with it, the recursive delete still runs outside the lock, since by then the tree only answers to an unguessable staging name.

### Evidence on the merged tree

```
$ npm run test:omp
EXIT=0    ℹ pass 114  ℹ fail 0    +    97 passing

$ npx mocha tests/unit/omp-*.test.js tests/omp-docker-fresh-only.test.js --timeout 30000
140 passing

$ npx mocha tests/unit/provider-session-capture.test.js \
            tests/unit/provider-session-restore.test.js \
            tests/unit/provider-session-reuse.test.js --timeout 30000
19 passing

$ npm test
EXIT=0                              # 2451 passing

$ npm run check:agent-cli-provider
ℹ pass 280   ℹ fail 0

$ npm run lint        → 0 errors (291 pre-existing repo-wide warnings)
$ npm run typecheck   → clean
```

The one caveat from my previous comment still stands and is unrelated to this PR: under a saturated parallel `npm test`, `tests/agent-stuck-recovery.test.js` can trip a 30s spawn-timeout expectation. It passes in isolation (`14 passing`) and `npm test` exits 0.

tomdps:
## Blocking issue fixed: cross-owner cleanup race

### The interleaving that could destroy a live session

Committed session `P` on partition `X`, two competing resumes `A` and `B`:

1. `A` and `B` are both inserted as `provisional` rows naming `X` (a resumed row exists *before* its transfer runs).
2. `transferOmpSessionOwnership(P → A)` wins: `P`'s record is cleared, `A` is `provisional` and holds the whole inherited lineage. **No row is `committed` at this point** — that is the normal state for the entire span of a resumed turn.
3. `B`'s transfer fails closed and its turn is retired to `cleanup-required`, holding a record that still names `X` but has `partitionIdentity: null`.
4. Cleanup driven by `B` consulted `findCommittedOwnersForPartition` → `[]`, and the descriptor-pinned identity check in the deleter is skipped when `partitionIdentity` is null → **`B` renamed and recursively deleted `A`'s live partition.**

### Fix

**Owner-fenced cleanup.** `findAuthoritativeOwnersForPartition` (new) returns every *other* row whose own valid record names the partition in an authoritative state — `provisional` **or** `committed`. Cleanup refuses if any exist. `cleanup-required` is deliberately excluded: it is the state of a turn that has already been retired and makes no claim, and treating it as authoritative would deadlock two retired rows against each other and strand the partition forever.

**Write fence closing the query-to-staging-rename race.** `src/omp-session-partition.js` now splits deletion into `stageOmpSessionPartitionForDeletion` (validate + atomic rename to `.zeroshot-deleting-<uuid>`) and `removeStagedOmpSessionPartition` (`rm -r`, guarded to only accept a `.zeroshot-deleting-*` child of an `omp-sessions/` root). Cleanup runs *re-read the row → check other claims → rename* inside one `BEGIN IMMEDIATE` task-store transaction, and the recursive removal after it. So a competing ownership write is serialized either wholly before the checks (seen by them) or wholly after the rename (it lands on a partition already gone and fails that turn closed, rather than costing a live turn its data) — and no unbounded `rm -r` is held under a write lock.

The fenced **row re-read** closes the same race from the caller's side: every surface iterates one `loadTasks()` snapshot, so a resume that transferred the partition away in the meantime leaves that snapshot describing a record the row no longer holds. The column is only ever written through `serializeOmpSessionOwnership`, whose output is canonical per record, so comparing stored bytes is an exact "still the same record" test.

### Tests added

| Test | File |
|---|---|
| `third-owner interleaving: the retired loser is the only row a committed-only fence can see` | `tests/unit/omp-session-ownership.test.js` |
| `preserves the partition of a resume that WON the transfer against a losing competitor` | `tests/unit/omp-session-cleanup.test.js` |
| `refuses a snapshot whose row transferred the partition away after it was loaded` (pre-stage concurrent transfer) | `tests/unit/omp-session-cleanup.test.js` |
| `refuses a stale snapshot even when no other row claims the partition` (isolates the re-read) | `tests/unit/omp-session-cleanup.test.js` |
| `stays idempotent across several retired rows naming one partition, without deadlocking` | `tests/unit/omp-session-cleanup.test.js` |

The ownership test pins the crux directly: in the third-owner state `findCommittedOwnersForPartition` returns `[]` while `findAuthoritativeOwnersForPartition` returns `[{taskId: winner, state: 'provisional'}]` — i.e. exactly why the committed-only fence let the deletion through.

### Corrected false claims

The "the transfer keeps exactly one row referencing the partition" / "never owned by two rows, never by none" statements were wrong and were what the old fence was built on. Corrected in `task-lib/omp-session-ownership.js`, `task-lib/omp-session-cleanup.js`, `AGENTS.md`, `docs/providers.md`, and a watcher-test assertion message: a partition has at most one **committed** owner, but regularly has **none** (the whole resumed turn) and is regularly named by **several rows at once**.

### Reconciled with current main (#909 merged)

Merged `origin/main` (`64fb06d`). Conflicts in `AGENTS.md`, `src/agent-cli-provider/provider-registry.ts`, and `tests/agent-cli-provider/omp-registry-capabilities.test.js` were resolved preserving both features: OMP now carries `dockerIsolation: true` (#909) **and** `sessionResume: true` (#866). These are independent — Docker stays fresh-only via `ZEROSHOT_OMP_SESSIONLESS`, so an isolated run allocates no partition, persists no ownership row, and launches `--no-session`. `tests/omp-docker-fresh-only.test.js` was updated from "sessionless for every OMP task" to the real post-#866 contract, and now also pins the positive side so a future change cannot quietly re-disable resume everywhere and still pass.

### Evidence

```
$ npx mocha tests/unit/omp-session-cleanup.test.js tests/unit/omp-session-ownership.test.js \
    tests/unit/omp-session-ownership-schema.test.js tests/unit/omp-resume-plan.test.js \
    tests/unit/omp-provider-session.test.js tests/unit/omp-session-verifier.test.js \
    tests/unit/omp-blob-root.test.js tests/omp-rpc-watcher.test.js
  162 passing (9s)                                    exit 0

$ npm run test:omp
  97 passing (9s)                                     exit 0

$ npm run lint
  ✖ 291 problems (0 errors, 291 warnings)             exit 0   # warnings all pre-existing

$ npm run typecheck
  tsc --noEmit && tsc --project tsconfig.cluster.json exit 0    # clean

$ npm test
  2460 passing (42s)
  18 pending                                          exit 0
```

Full suite run three times on the pushed tree; runs 1 and 3 were clean end to end. Run 2 showed five transient failures in `tests/agent-stuck-recovery.test.js` (signal/timing assertions) on a machine under heavy concurrent load — mocha still exited 0 on all three runs, and that file passes 3/3 in isolation. It is untouched by this change, which adds no process or signal code.

PR #909, opened 2026-08-02
feat: feat(omp): add Docker distribution and broker auth

### LINKED ISSUE #868: feat(omp): add Docker distribution and broker auth

tomdps:
Closes #868


tomdps:
Blocking review; please fix before merge.

1. Greptile's P1 is valid: `imageForProvider()` appends a suffix after a digest, producing an invalid image reference. Handle digest-pinned and registry-port/tag references explicitly, include the full base reference in the derived cache identity, and add focused cases.
2. The effective auth plan stores `forwardedEnv[name] = true`. An explicit forced-empty entry such as `dockerEnvPassthrough: ["OPENAI_API_KEY="]` is therefore treated as authenticated even though Docker receives an empty value. Preserve actual values in the internal plan, reject empty/whitespace-only values, validate a broker URL as a usable http(s) URL, and never print values.
3. Explicitly opted-in OMP credentials cannot work. `docker.envAuth.requireOneOf` only accepts the four automatic key/token names, so a user who explicitly forwards `OPENROUTER_API_KEY`, `ANTHROPIC_OAUTH_TOKEN`, or another registry-known OMP credential still fails. The issue requires non-allowlisted custom env/path credentials to require explicit passthrough/mount—not to remain unusable. Accept a registry-known credential only when it is present in the effective explicit plan; preserve the five-name automatic allowlist and broker pair rule.
4. Path credentials need both the explicit container-path env and a matching explicit mount. Do not count a host path merely because it exists, and add a negative test for an explicit path env without its mount. Conversely, the current remediation derives `mountHint` from OMP's `credentialPaths[0]` and recommends mounting `~/.omp`, directly contradicting the requirement never to mount/copy the host OMP auth store. Env-only provider remediation must use a generic custom path example and prefer the broker.
5. Parse Buildx `Platforms:` as exact comma-delimited platform tokens; `includes("linux/amd64")` accepts only-advertised variants such as `linux/amd64/v2`. Add exact-match tests.
6. Move the no-secret/malformed effective-plan check before `_prepareIsolatedWorkspace()` and `_removeContainerByName()` so failed auth has no workspace/container effects. The platform probe already has this ordering; auth should too.
7. Keep the OMP five-name automatic allowlist negative coverage, and add tests for forced-empty values, whitespace, malformed broker URL, explicit custom API key success, path env without mount failure, path env plus mount success, digest/tag/registry-port images, and exact Buildx token matching. Run `npm run test:omp` and the focused Docker suites, then reply with evidence.

tomdps:
All seven items are fixed on `zeroshot/lunar-core-78` (`eb12db7`). Evidence below.

### 1. Greptile P1 — invalid derived image references

`IsolationManager.parseImageReference` splits `[registry[:port]/]name[:tag][@digest]`; the only ambiguity is `:`, resolved by position (before the last `/` → registry port, after it → tag). `imageForProvider` now derives the variant from the **name** component only, so a digest or tag can no longer be carried into the suffix. The hash covers the **full** base reference alongside install command and platform, so two pins of one base name never collide on a cached tag.

| base reference | derived |
| --- | --- |
| `registry.example/base@sha256:aaa…` | `registry.example/base-omp-<hash>` |
| `zeroshot-cluster-base:v2` | `zeroshot-cluster-base-omp-<hash>` |
| `registry.example:5000/base` | `registry.example:5000/base-omp-<hash>` (port kept) |
| `registry.example:5000/team/base:v2@sha256:aaa…` | `registry.example:5000/team/base-omp-<hash>` |

`tests/unit/provider-docker-image.test.js` — 5 digest/tag/port derivation cases each validated by an independent hand-split reference oracle (`assertValidImageReference`, deliberately not the production parser), 3 cache-identity cases (differing digests / differing tags / tagged vs untagged all yield distinct tags), a stability case, a claude digest-passthrough case, and 6 `parseImageReference` unit cases.

### 2. Actual values in the effective plan

`_buildCredentialPlan` stores `forwardedEnv[name] = <actual value>`. `isUsableEnvValue` rejects absent/non-string/empty/whitespace-only. `docker.envAuth.requireUrl: ['OMP_AUTH_BROKER_URL']` requires an absolute http(s) URL (per OMP v17.2.1 `docs/environment-variables.md`: it is the broker's base URL, e.g. `https://broker.tailnet:8765`, and OMP hard-errors on a broker URL with no resolvable token). A malformed URL is a hard defect no other credential compensates for — `validateProviderEnvAuth` returns `{ok, satisfiedOneOf, malformed[], message}` and the caller gates on `malformed.length === 0`. Every message names variables only; the test suite asserts the exact secret/URL substrings are absent.

The forced-empty case is proven end-to-end, not just at the unit boundary: the test asserts `plan.forwardedEnv.OPENAI_API_KEY === ''` **and** `plan.args.includes('OPENAI_API_KEY=')` (Docker really does receive it) **and** that `createContainer` still rejects — exactly the gap a presence flag left open.

### 3. Explicitly opted-in registry-known credentials now work

`_classifyForwardedCredentials` accepts a credential when it is a registry-known `credentialEnvKeys` name **and** either on the automatic allowlist (`requireOneOf`) or present in the *explicit* plan. `_collectDockerEnvVars` now computes explicitness from the user's own config (`dockerEnvPassthrough` + explicitly listed presets) **before** `_withActiveProviderPreset` auto-activates the running provider's preset, so "automatic" and "explicit" are genuinely distinguished. `OPENROUTER_API_KEY` via `dockerEnvPassthrough` succeeds; the same key in the host env alone still fails and is never forwarded. The five-name automatic allowlist and the broker pair rule are unchanged (`requireOneOf` stays the four automatic *credentials* — `OMP_AUTH_BROKER_URL` is a locator that authenticates nothing alone, which is why it is in `requireTogether`, and a new registry test asserts every `requireOneOf` name is both a real adapter credential and automatically forwarded, so the entry cannot drift into being unsatisfiable).

### 4. Path credentials + safe remediation

A forwarded credential whose value is an absolute container path counts only when an **explicitly configured** mount actually provides that path (`pathContains(mountContainerPath, value)`). Host-side existence is never consulted. Negative coverage: explicit path env with no mount; a mount at a *different* container path; and a value that is a real, existing **host** path — all reject with `… points at a container path that no explicit --mount provides`. Positive coverage: exact-file mount, and a directory mount containing the credential.

`_credentialRemediation` no longer derives a hint from `credentialPaths[0]` for env-only providers. It prefers the broker (`Prefer the auth broker (OMP_AUTH_BROKER_URL + OMP_AUTH_BROKER_TOKEN) so host refresh tokens never cross.`) and uses a generic example (`--mount /host/path/to/credential:/home/node/credential:ro`). A test asserts the message never contains `credentialPaths[0]` and never matches `--mount ~`. Mount-based providers keep their existing concrete hint.

### 5. Exact Buildx platform tokens

`parseBuildxPlatforms` splits the `Platforms:` line on commas, trims, strips buildx's trailing preferred-marker `*`, and collects **every** builder node's line. `linux/amd64/v2` is a variant, not `linux/amd64` — a builder advertising only variants now correctly fails. `assertPlatformSupported` takes an injectable probe, so the whole decision table is tested directly: native `amd64`, native `x86_64`, emulated arm64 with exact `linux/amd64`, emulated arm64 with **only** variants (rejected), no buildx (rejected), non-Linux server (rejected), and `docker info` failure. The binfmt remediation is now derived from the required arch rather than hardcoded.

### 6. Pre-effect auth ordering

`createContainer` computes the read-only plan and validates it **before** `_removeContainerByName()` and `_prepareIsolatedWorkspace()`, matching the platform probe's ordering. The test wraps both methods and asserts neither ran, no container spawned, no cluster config dir created, and no `zeroshot-isolated/<clusterId>` directory left on disk — plus a positive control proving both effects still run once auth is satisfied.

### 7. Verification

```
npm run test:omp                → EXIT 0 — ℹ tests 108 / pass 108 / fail 0, plus 78 passing (mocha)
focused Docker suites           → EXIT 0 — 207 passing
  tests/unit/docker-config.test.js
  tests/unit/docker-provider-preset.test.js
  tests/unit/provider-docker-image.test.js
  tests/omp-docker-auth.test.js
  tests/omp-docker-fresh-only.test.js
  tests/omp-isolation-capability.test.js
  tests/unit/worktree-claude-config.test.js
npm test                        → EXIT 0 — 0 failures
npm run lint                    → EXIT 0 — 266 problems (0 errors, 266 warnings, all pre-existing)
npm run typecheck               → EXIT 0
npm run check:agent-cli-provider:ci → EXIT 0 — ℹ tests 272 / pass 272 / fail 0
```

New/updated coverage: forced-empty (`VAR=`) values, whitespace-only values, malformed and non-http(s) broker URLs, malformed URL alongside a valid key, explicit custom API key success plus its host-env-only failure, path env without mount, path env with exact-file and with directory mount, path env with a mismatched mount, host-path-that-merely-exists, digest/tag/registry-port image derivation and cache identity, exact Buildx token matching, the pre-effect ordering assertion, and a regression guard that a bare `true` presence flag is no longer accepted as a credential. The five-name automatic allowlist negative coverage is retained unchanged.

`AGENTS.md`'s OMP Docker section is updated for all of the above. No paid model request was made.

PR #934, opened 2026-08-04
feat(omp): benchmark through pinned SDK

tomdps:
## Summary
- make omitted OMP transport execute through the pinned OMP SDK and Bun runtime
- preserve explicit RPC compatibility while enforcing strict terminal yield, schema, model, and cleanup evidence
- publish and attest the complete SDK sidecar runtime closure for deterministic installs

## Verification
- npm run check
- npm test (2602 passing, 18 pending)
- npm run check:agent-cli-provider:ci (292 passing)
- npm run test:omp (102 passing)
- npm run release:preflight
- npm pack --dry-run --json


tomdps:
The non-Linux point is intentional, not a fallback bug. Omitted transport now means the pinned SDK lane; explicit `transport: rpc` remains the compatibility lane. The published SDK runtime closure is deliberately attested only for `linux:x64` and `linux:arm64`, matching the Harbor benchmark environment. Falling back implicitly on macOS/Windows would violate exact execution identity and make omitted transport host-dependent. The provider therefore fails closed on unsupported hosts and reports the explicit RPC remediation.

PR #938, opened 2026-08-05
fix: patch brace-expansion in hosted OECP image

anupamme:
## Summary

Patch the nested `brace-expansion` installation used by `@earendil-works/pi-coding-agent` from 5.0.6 to 5.0.9 in the private hosted OECP image.

## Security

- Fixes CVE-2026-14257, addressed in the 5.x line by 5.0.8.
- Also incorporates 5.0.9, which addresses CVE-2026-69152 affecting 5.0.8.
- Affected path: `docker/zeroshot-oecp/package-lock.json` through Pi's nested `minimatch` dependency.
- Presence was confirmed; runtime reachability was not.

## Root cause

`@earendil-works/pi-coding-agent@0.80.3` ships its own `npm-shrinkwrap.json`. `npm ci` respects that shrinkwrap, so a root override alone does not replace the installed nested 5.0.6 package.

The image build therefore installs the exact patched release with lifecycle scripts disabled, copies it over Pi's shrinkwrap-owned nested package, removes the temporary top-level copy, and asserts the final installed version.

## Changes

- `docker/zeroshot-oecp/package.json`: declare the 5.0.9 override.
- `docker/zeroshot-oecp/package-lock.json`: record the patched resolution and integrity.
- `docker/zeroshot-oecp/Dockerfile`: replace the shrinkwrap-owned installation and assert version 5.0.9.
- `docker/zeroshot-oecp/build-manifest.json`: regenerate image input digests.

The unpublished hosted OECP runtime remains private and otherwise unchanged.

### DIFF
--- a/docker/zeroshot-oecp/Dockerfile
+++ b/docker/zeroshot-oecp/Dockerfile
@@ -10,9 +10,12 @@ FROM docker.io/library/node:22-bookworm-slim@sha256:f32b81066cde10a75dbac9664609
 WORKDIR /opt/node-runtime
 COPY docker/zeroshot-oecp/package.json docker/zeroshot-oecp/package-lock.json ./
 RUN npm ci --omit=dev --ignore-scripts --no-audit --no-fund \
-    && npm install brace-expansion@5.0.8 --no-save --ignore-scripts --no-audit --no-fund \
+    && npm install brace-expansion@5.0.9 --no-save --ignore-scripts --no-audit --no-fund \
     && cp -r node_modules/brace-expansion/. node_modules/@earendil-works/pi-coding-agent/node_modules/brace-expansion/ \
-    && rm -rf node_modules/brace-expansion
+    && rm -rf node_modules/brace-expansion \
+    && node -e "const pkg = require( \
+      './node_modules/@earendil-works/pi-coding-agent/node_modules/brace-expansion/package.json'); \
+      if (pkg.version !== '5.0.9') throw new Error('expected brace-expansion@5.0.9, got ' + pkg.version)"
 
 FROM docker.io/instructure/tini:v0.19.0@sha256:0f9ccf2e54d010735ad691ebe670e256a3820fe8b92c28101194c68ee1216092 AS tini
 
--- a/docker/zeroshot-oecp/Dockerfile
+++ b/docker/zeroshot-oecp/Dockerfile
@@ -9,7 +9,10 @@ RUN cargo build --locked --release -p zeroshot-rust --bin zeroshot-oecp-server
 FROM docker.io/library/node:22-bookworm-slim@sha256:f32b81066cde10a75dbac96646099533316d94bac4150c55da1636e1f0ffdc46 AS node-deps
 WORKDIR /opt/node-runtime
 COPY docker/zeroshot-oecp/package.json docker/zeroshot-oecp/package-lock.json ./
-RUN npm ci --omit=dev --ignore-scripts --no-audit --no-fund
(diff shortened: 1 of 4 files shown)


tomdps (changes requested):
Thanks for jumping on a newly published advisory, especially on your first contribution here. The affected lockfile really does contain brace-expansion 5.0.6 and 5.0.8 is the correct patched release, but this revision cannot build as written.

The manifest overrides brace-expansion to 5.0.8 while the lockfile still resolves 5.0.6, so the image's npm ci step sees an inconsistent manifest/lock pair. The lockfile also drops unrelated libc metadata and changes the pi-ai bin path; please regenerate it with the repository's expected npm version and keep the diff limited to the security resolution.

For this tree there is one affected 5.x installation, through minimatch, so please also narrow the title/body rather than claiming four major-version upgrades. A complete revision should show:

- docker/zeroshot-oecp/package-lock.json actually resolving brace-expansion 5.0.8;
- npm ci succeeding in docker/zeroshot-oecp;
- the hosted OECP image build/smoke succeeding; and
- a Trivy rescan no longer reporting CVE-2026-14257.

Happy to re-review once those are in place.

anupamme:
Thanks for the detailed review. Here's what I found and fixed:

- Root cause: @earendil-works/pi-coding-agent@0.80.3 ships its own npm-shrinkwrap.json (hasShrinkwrap: true in the lockfile). npm ci respects that shrinkwrap for the package's subtree, so the overrides field in package.json couldn't actually replace
- brace-expansion@5.0.6 in the installed tree: the vulnerable version was always installed regardless.

  What this revision does:

  - package-lock.json: lockfile now records brace-expansion@5.0.8 with the correct resolved URL and integrity hash, and the root entry carries the overrides declaration. The diff is limited to the security resolution only; no libc metadata changes, no unrelated bin path changes. npm ci succeeds locally.
  - Dockerfile: after npm ci, installs brace-expansion@5.0.8 at the top level, copies it over the shrinkwrap-locked copy under @earendil-works/pi-coding-agent/node_modules/brace-expansion/, then removes the transient top-level copy. This is necessary because no upstream pi-coding-agent release currently ships >=5.0.8 in its shrinkwrap.
  - build-manifest.json: regenerated against current inputs; node scripts/hosted-oecp-image.js check passes cleanly.

I wasn't able to run the Docker build/smoke locally (no Docker available in this environment), so I'd ask CI to confirm items 3 and 4. Happy to address any follow-up.

tomdps:
Thanks — your shrinkwrap diagnosis was correct. We verified that the revised image builds and that the replacement is applied as intended.

I also confirmed that our PR CI does not currently run the hosted-image smoke or Trivy, and the same smoke is failing on current `main`. That validation gap is ours, not something you need to resolve.

Since your update, `brace-expansion@5.0.9` was released to address another advisory affecting 5.0.8. We’re going to apply that small version update and an installed-version assertion directly to this branch, tidy the PR title/body, and complete the validation before merging. No further action is needed from you.

Thanks again for tracking down the shrinkwrap behavior and getting this into a buildable state.

tomdps (approved):
Verified the maintainer follow-up on the updated branch:

- `npm ci` accepts the manifest/lock pair.
- Pi resolves and executes `brace-expansion@5.0.9` after the shrinkwrap replacement.
- The generated OECP manifest check passes.
- The hosted OECP image builds and its inspection contract passes.
- The final image resolves the nested package as 5.0.9.
- A targeted Trivy filesystem scan reports no `brace-expansion` findings.

The hosted smoke failure is also present on current `main` and is not attributable to this change. Approved.

PR #979, opened 2026-08-10
feat(cli): add atomic foreground benchmark results

tomdps:
## Summary
- add a foreground-only `--result-file` contract with atomic result and telemetry receipts
- make terminal settlement, cancellation, and exit classification explicit and fail closed
- use the pipe watcher for benchmark execution so remote PTY tail races cannot drop structured terminal output
- add deterministic crash, signal, large-output, liveness, and terminal-tail regressions

## Why
Benchmarks needs one authoritative, bounded, programmatic result boundary instead of reconstructing terminal truth from detached process state, status polling, and full ledger exports. This also fixes the Modal-only PTY tail-loss race reproduced by the zero-token qualification gate.

## Validation
- 2788 unit tests passed; 18 pending
- 33 E2E tests passed
- 12 MiB terminal-output stress passed
- zero-token Modal foreground Harbor gate passed with reward 1.0
- zero-token Docker success and provider-failure gates passed
- lint/typecheck/package/release preflight passed
- `npm run opcore:check`, direct Opcore changed, and Opcore Zero passed

Tracks the-open-engine/benchmarks#283 and the-open-engine/benchmarks#286.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -74,6 +74,11 @@ const {
 } = require('../lib/start-cluster');
 const { requirePreflight } = require('../src/preflight');
 const {
+  exitCodeForResult,
+  isForegroundStatusSettled,
+  writeForegroundResult,
+} = require('../src/foreground-benchmark-run');
+const {
   createUnsupportedProviderCapabilityError,
   serializeTaskStartupError,
 } = require('../src/task-startup-error');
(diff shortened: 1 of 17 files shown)


tomdps:
Addressed the execution-context finding in 3ad7d48. `--result-file` now establishes `ZEROSHOT_TASK_EXECUTION_CONTEXT=benchmark` inside the CLI before cluster construction, so direct programmatic consumers do not rely on an outer wrapper to select the close-delimited pipe watcher. The real foreground E2E starts with all `ZEROSHOT_*` variables scrubbed and now asserts the persisted provider task has `attachable = 0`.

Issue #1009, opened 2026-08-12
[BUG] js-yaml dependency is required at runtime


s0kil:
## Description

A clear description of the bug.

## Steps to Reproduce
Run `zeroshot run "Add a --json flag with tests"`

## Expected Behavior
Should finish and auto-stop cluster

## Actual Behavior
```bash
  Failed to auto-stop cluster zen-spire-47: Cannot find module 'js-yaml'
  Require stack:
  - /home/danielsokil/.local/share/pnpm/store/v11/links/@the-open-engine/zeroshot/6.39.0/cf6f922de8de2c501e09a216d6f8820a450ddb8cc82fc6f0c6040b9c14f5a36c/node_modules/@the-open-engine/zeroshot/lib/compose-utils.js
  - /home/danielsokil/.local/share/pnpm/store/v11/links/@the-open-engine/zeroshot/6.39.0/cf6f922de8de2c501e09a216d6f8820a450ddb8cc82fc6f0c6040b9c14f5a36c/node_modules/@the-open-engine/zeroshot/src/orchestrator.js
  - /home/danielsokil/.local/share/pnpm/store/v11/links/@the-open-engine/zeroshot/6.39.0/cf6f922de8de2c501e09a216d6f8820a450ddb8cc82fc6f0c6040b9c14f5a36c/node_modules/@the-open-engine/zeroshot/cli/index.js
```

$ zeroshot --version
6.39.0

s0kil:
I've tested with npm, node version `v22.23.2`, same issue as with pnpm

PR #1010, opened 2026-08-12
fix: ship js-yaml as a runtime dependency

s0kil:
## Problem

Every cluster that finishes in a worktree crashes during auto-stop:

```
Failed to auto-stop cluster zen-spire-47: Cannot find module 'js-yaml'
Require stack:
- .../@the-open-engine/zeroshot/lib/compose-utils.js
- .../@the-open-engine/zeroshot/src/orchestrator.js
- .../@the-open-engine/zeroshot/cli/index.js
```

`lib/compose-utils.js` (built from `src/legacy-lib/compose-utils.ts`) requires `js-yaml` at runtime, but `js-yaml` was declared as a **devDependency**. `npm-shrinkwrap.json` therefore marked it `"dev": true`, so no consumer install ever received it.

Flat npm trees could mask this by hoisting `js-yaml` from an unrelated package into top-level `node_modules`. pnpm's isolated store has no such accident, so the failure surfaced there first — but the dependency was never actually declared for any install.

A second defect turned a cosmetic gap into a fatal one: both call sites `require('../lib/compose-utils')` **outside** their `try` blocks, so a module-resolution failure escaped the "best-effort, silently ignores failures" contract and aborted cluster shutdown.

## Changes

- Move `js-yaml` from `devDependencies` to `dependencies`; regenerate `package-lock.json` / `npm-shrinkwrap.json` (clears `dev: true` on `js-yaml` and its `argparse` dependency).
- Update `checkScriptDependencies()` in `scripts/rust-distribution.js`, which asserted the *devDependency* placement and would have failed the release integrity gate.
- Load `lib/compose-utils` inside the teardown `try` blocks in `src/orchestrator.js` and `src/isolation-manager.js` so compose cleanup stays best-effort. Orchestrator logs the skip reason; worktree removal still proceeds.

## Tests

- `tests/unit/published-runtime-dependencies.test.js` (new): scans every published CommonJS file (respecting the `files` allow/deny list) and asserts each bare `require()` resolves to a declared runtime or optional dependency. This is the guard that would have caught the bug at commit time.
- `tests/worktree-compose-cleanup.test.js`: two regression tests covering an unloadable `lib/compose-utils` — teardown is skipped without throwing, and worktree removal still runs.
- `tests/unit/rust-release-workflow-dependencies.test.js`: updated for the new dependency placement.

## Verification

```
npm test    # 2903 passing, 18 pending
npm run lint         # 0 errors
npm run typecheck    # clean
npm ls --omit=dev js-yaml
└── js-yaml@4.2.0    # previously unresolvable in a production install
```

3 pre-existing failures in `tests/unit/opcore-introduced-gate.test.js` reproduce identically on unmodified `main` in this environment (local `rust.function-metrics` reports `unsupported_request`) and are unrelated to this change.


s0kil:
Closes https://github.com/the-open-engine/zeroshot/issues/1009

tomdps (approved):
Confirmed the published-runtime regression and reviewed the fix end to end. js-yaml is now shipped at patched 4.3.1, both lockfiles remain synchronized, teardown failures are diagnosed without blocking worktree removal, and focused plus full CI—including production audit and Ubuntu/macOS install checks—passes on this exact head.

PR #1017, opened 2026-08-14
fix(ship): count staged validation approvals in the git-pusher handoff

EivMeyer:
## Scope — read this first

**This is not a live bug on `main`.** Nothing on `main` publishes
`STAGE_<n>_VALIDATION_RESULT`, so the handoff cannot currently fail this way.
It is a forward-compatibility fix: it becomes a real bug the moment a topology
that validates in stages is used, which is exactly what
`eivind/topology-generator` produces.

I hit it there, not on main. Flagging that because the first version of this
description implied otherwise.

## The problem

`git-pusher` commits, pushes and opens the PR in ship mode, and only fires once
every validator has approved. It established that by counting `VALIDATION_RESULT`
messages:

```js
if (latestByValidator.size < validators.length) return false;
```

A staged topology publishes `STAGE_<n>_VALIDATION_RESULT` for every stage
except the last, so those validators never counted, the size check could never
be satisfied, and `git-pusher` stayed `idle` forever.

Observed on a real two-stage cluster, both approving:

```
STAGE_1_VALIDATION_RESULT  <- toolchain-gate           ← not counted
VALIDATION_RESULT          <- cancel-behaviour-audit   ← counted

latestByValidator.size = 1  <  validators.length = 2   →  no ship
```

The change was built, every validator approved it, and it was abandoned
uncommitted in the worktree while the run reported retries exhausted.

## The fix

Match `^(?:STAGE_[0-9]+_)?VALIDATION_RESULT$`.

Two deliberate details:

- **Exact, not suffix.** My first attempt used `/(?:^|_)VALIDATION_RESULT$/`,
  which also matches `QUICK_VALIDATION_RESULT` and `HEAVY_VALIDATION_RESULT` —
  both present in this repo as tiers of the same check. That would have let a
  cheap pre-check satisfy the ship gate, which is strictly worse than the bug:
  not shipping is recoverable, shipping unverified work is not.
- **`[0-9]` rather than `\d`.** `SHARED_TRIGGER_SCRIPT` is a template literal,
  so the backslash is consumed before the script is compiled — `\d` silently
  became `d` and the staged topic stopped matching at all.

## Verification

The test compiles the trigger through `vm` exactly as the runtime does, rather
than reimplementing its logic.

- against `main`: **4 passing, 1 failing**
- with this branch: **5 passing**

Covered: flat topology still hands off; a staged approval counts; a rejected
stage still blocks; a validator that never reported still blocks; and neither
QUICK nor HEAVY tier can satisfy the gate.

## Trade-off worth a reviewer's eye

The query drops its indexed `topic` filter and filters in JS instead, so it now
scans messages since the last `IMPLEMENTATION_READY` rather than hitting the
topic index. That window is the tail of a run, so it is bounded — but if you'd
rather keep the index, the alternative is two targeted queries, or having the
topology generator emit a terminal `VALIDATION_RESULT` per validator.

### DIFF
--- a/src/agents/git-pusher-template.js
+++ b/src/agents/git-pusher-template.js
@@ -216,7 +216,14 @@ function assertRequiredQualityGatesPass(messages) {
 const requiredQualityGatesForHandoff = getRequiredQualityGates();
 if (validators.length === 0 && requiredQualityGatesForHandoff.length === 0) return true;
 
-const results = ledger.query({ topic: 'VALIDATION_RESULT', since: lastPush.timestamp });
+// A generated topology validates in stages, and every stage but the last
+// publishes STAGE_<n>_VALIDATION_RESULT rather than VALIDATION_RESULT. Counting
+// only the plain topic misses every earlier-stage validator, so
+// latestByValidator can never reach validators.length and the handoff never
+// fires: the work is built, approved by all of them, and then never shipped.
+const results = ledger
+  .query({ since: lastPush.timestamp })
+  .filter((msg) => typeof msg.topic === 'string' && /(?:^|_)VALIDATION_RESULT$/.test(msg.topic));
 if (results.length === 0) return false;
 
 const validatorIds = new Set(validators.map((v) => v.id));
--- a/src/agents/git-pusher-template.js
+++ b/src/agents/git-pusher-template.js
@@ -216,14 +216,19 @@ function assertRequiredQualityGatesPass(messages) {
 const requiredQualityGatesForHandoff = getRequiredQualityGates();
 if (validators.length === 0 && requiredQualityGatesForHandoff.length === 0) return true;
 
-// A generated topology validates in stages, and every stage but the last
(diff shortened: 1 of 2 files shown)


EivMeyer:
Closing: this patches the wrong layer.

`git-pusher`'s contract is "every validator publishes VALIDATION_RESULT". The
topology generator broke that by hijacking the validation topic to chain stages,
so the fix belongs there, not in core.

Tolerating it here costs two things I am not willing to spend: the topic match
has to loosen (and this repo has QUICK_/HEAVY_VALIDATION_RESULT as tiers of the
same check, so a suffix match would let a cheap pre-check satisfy the ship gate),
and the query loses its topic index.

The generator can meet the existing contract instead — publish VALIDATION_RESULT
from every stage and carry the stage in the payload, triggering the next stage on
that field rather than a renamed topic.

Written up in #1018, including the test, which is worth keeping wherever the fix
lands.

PR #1021, opened 2026-08-16
docs: add zeroshot task spec and review skills

s0kil:
Add two Agent Skills (agentskills.io format) that teach coding agents how to write task files Zeroshot can execute and independently verify.

zeroshot-task-spec covers the mechanics that are not documented elsewhere and that silently degrade a run when missed:

- only .md/.markdown are read as files; any other extension makes the literal path string the task body
- the first H1 becomes the cluster title and the PR title
- the zeroshot-command-proofs fenced block is the only machine-parsed hook in a task file, and each entry becomes a fail-closed handoff gate under --pr/--ship
- acceptance criteria shaped {id, criterion, verification, priority} survive the planner handoff intact, while checkbox lists get paraphrased first
- a criterion whose verification is not runnable lands in CANNOT_VALIDATE, which is treated as a pass, so it cannot fail
- the conductor's tie-break bias runs downward, so an unstated risk category buys a cheaper workflow than the change warrants

zeroshot-task-review audits a draft before the run and emits a verdict, blocking issues each paired with replacement text, and a corrected criteria table.

Content is derived from cluster-templates/conductor-bootstrap.json, cluster-templates/base-templates/, src/command-proofs.ts, src/config-router.ts, src/input-helpers.ts and src/agents/git-pusher-template.js rather than from prose docs.

Developed against baseline runs: three agents wrote specs without the skill and produced three different structures, none declaring command proofs, none anchoring classification, and none using the criteria contract. The same scenarios with the skill produced all three consistently.

## Summary

Brief description of changes.

## Related Issues

Fixes #

## Changes Made

-

## Testing

How was this tested?

## Checklist

- [ ] Tests pass (`npm test`)
- [ ] Documentation updated (if needed)
- [ ] Follows commit guidelines

### DIFF
--- a/skills/zeroshot-task-review/SKILL.md
+++ b/skills/zeroshot-task-review/SKILL.md
@@ -1,0 +1,115 @@
+---
+name: zeroshot-task-review
+description: Use when a Zeroshot task or plan file has been drafted and needs checking before `zeroshot run`, when reviewing someone else's Zeroshot spec, or when deciding whether a task file is ready to hand to Zeroshot.
+license: MIT
+metadata:
+  author: the-open-engine
+  version: '1.0'
+---
+
+# Reviewing Zeroshot Task Specs
+
+## Overview
+
+A Zeroshot spec fails in ways that only show up an hour into a run: a criterion
+nothing can check, an unstated risk that bought the cheap workflow, a deferred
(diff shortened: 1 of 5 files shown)


mkceichelbeck:

Issue #70, opened 2026-01-15
feat: Windows native support (socket IPC + executable resolution)


EivMeyer:
## Summary

Windows native support is completely broken. Two interconnected issues block all task execution:

- #63: Unix sockets fail (`.sock` files not supported on Windows)
- #57: `spawn zeroshot ENOENT` (bare command resolution fails)

### Socket IPC (#63)

`attach-server.js` uses Unix domain sockets (`~/.zeroshot/sockets/task.sock`). Windows needs named pipes (`\\.\pipe\zeroshot-task`).

Node.js `net` module **already supports** Windows named pipes natively - just needs different path format.

### Executable Resolution (#57)

`getClaudeTasksPath()` returns bare `'zeroshot'` string. Windows `spawn()` without shell can't resolve bare commands - needs full path or `shell: true`.

### 1. Socket abstraction (`src/attach/socket-discovery.js`)

Add platform-aware path construction:

```javascript
const IS_WINDOWS = process.platform === 'win32';

function getSocketPath(id, type = 'task') {
  if (IS_WINDOWS) {
    return `\\\\.\\pipe\\zeroshot-${type}-${id}`;
  }
  return path.join(SOCKET_DIR, `${id}.sock`);
}
```

Update all path functions: `getTaskSocketPath`, `getAgentSocketPath`, `listAttachableTasks`

### 2. Attach server fixes (`src/attach/attach-server.js`)

- Skip `fs.chmod()` on Windows (named pipes handle permissions differently)
- Skip `fs.unlink()` cleanup on Windows (OS handles pipe cleanup)

### 3. Spawn fix (`src/agent/agent-task-executor.js`)

Change `getClaudeTasksPath()` to resolve executable:

```javascript
function getClaudeTasksPath() {
  if (process.platform === 'win32') {
    // npm global installs create .cmd wrapper on Windows
    const result = execSync('where zeroshot', { encoding: 'utf8' }).trim().split('\n')[0];
    return result;
  }
  return 'zeroshot';
}
```

Or simpler: add `shell: true` to spawn options.

### 4. Tests

- Unit tests for socket path generation (mock `process.platform`)
- Unit tests for executable resolution

## Files to Modify

| File | Changes |
|------|---------|
| `src/attach/socket-discovery.js` | Platform-aware paths, pipe listing |
| `src/attach/attach-server.js` | Conditional chmod/unlink |
| `src/agent/agent-task-executor.js` | Resolve executable path |

## Testing

[...]

Issue #63, opened 2026-01-14
[BUG] Windows Native Version Not Working


LavelleUp:
## Description

Zeroshot tasks fail immediately on Windows native (not WSL) due to Unix socket IPC not being supported. The attach server tries to create a `.sock` file which Windows doesn't handle properly.

## Steps to Reproduce

1. Install zeroshot on Windows native: `npm install -g @covibes/zeroshot` (or from dev branch)
2. Run any task: `zeroshot task run "any task description"`
3. Task spawns but immediately fails

## Expected Behavior

Task should spawn and run Claude agent successfully.

## Actual Behavior

Task spawns but immediately dies with socket permission error:

```
Error: listen EACCES: permission denied C:\Users\xxx\.zeroshot\sockets\vivid-cosmos-18.sock
```

## Environment

- OS: Windows 10 (10.0.19045.6456) - Native, not WSL
- Node.js version: v24.4.1
- npm version: 11.4.2
- zeroshot version: 5.2.1 (dev branch, commit 5a3078c)
- Docker version: N/A (not using --docker)

## Logs

```
Task: vivid-cosmos-18

Status:     stale (process died)
Created:    2026-01-12T15:03:59.802Z
Updated:    2026-01-12T15:03:59.803Z
PID:        N/A
Exit Code:  N/A
Session:    N/A

Log file contents:
Error: listen EACCES: permission denied C:\Users\xxx\.zeroshot\sockets\vivid-cosmos-18.sock
```

## Analysis

The `attach-server.js` uses Node's `net.createServer()` with Unix socket paths (`.sock` files). Windows doesn't support Unix domain sockets in the filesystem the same way Linux/macOS do.

**Potential fix**: Implement Windows named pipes (`\\.\pipe\pipename`) as an alternative IPC mechanism when `process.platform === 'win32'`.

## Notes

- The recent dev branch Windows fixes (commit adf599c for `where.exe` command detection, commit f5bfb0d for CPU spin-wait) work correctly
- The `zeroshot --help` and `zeroshot --version` commands work fine
- Only the actual task execution fails due to socket creation

tomdps:
Status review (2026-07-29):
- Current main blocks native Windows in `runPreflight()` (`src/preflight.js`) before task/attach startup; its diagnostic also states that Windows support, including WSL, is deferred.
[...]

Issue #57, opened 2026-01-13
[BUG] Error: spawn zeroshot ENOENT


matvejs16:
## Description

Agent error with any prompt

## Steps to Reproduce

(Gemini-CLI default agent)
1. Run `zeroshot run "text prompt"`

## Expected Behavior

Agent will work on task

## Actual Behavior

Getting error:
```bash
────────────────────────────────────────────────────────────
junior-conductor | ⚡ ISSUE_OPENED → task #1 (null)
junior-conductor | • PROCESS_SPAWNED

────────────────────────────────────────────────────────────
junior-conductor | 14:52:40 🔴 AGENT ERROR
junior-conductor | Task execution failed after 1 attempts: spawn zeroshot ENOENT
junior-conductor | Error: spawn zeroshot ENOENT
junior-conductor |     at ChildProcess._handle.onexit (node:internal/child_process:285:19)
junior-conductor |     at onErrorNT (node:internal/child_process:483:16)
junior-conductor |     at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
────────────────────────────────────────────────────────────
```

### System:
 - OS: Windows 11 10.0.26200
 - CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11700K @ 3.60GHz
 - Memory: 25.83 GB / 31.87 GB
### Binaries:
 - Node: 22.16.0 - C:\Program Files\nodejs\node.EXE
 - npm: 11.7.0 - C:\Program Files\nodejs\npm.CMD
 - pnpm: 8.9.2 - C:\Users\matve\AppData\Local\pnpm\pnpm.EXE
 - bun: 1.3.5 - C:\Users\matve\.bun\bin\bun.EXE
### npmGlobalPackages:
 - @covibes/zeroshot: 5.3.0
 - @google/gemini-cli: 0.23.0

## Logs

```
✓ Preflight checks passed
Provider override: gemini (all agents)
Starting steady-spire-84
Config: conductor-bootstrap
Ctrl+C to stop following (cluster keeps running)


================================================================================
🔴 TASK EXECUTION FAILED - AGENT: junior-conductor (Attempt 1/1)
================================================================================
Error: spawn zeroshot ENOENT
================================================================================


================================================================================
🔴🔴🔴 MAX RETRIES EXHAUSTED - AGENT: junior-conductor 🔴🔴🔴
================================================================================
All 1 attempts failed
Final error: spawn zeroshot ENOENT
[...]

PR #1034, opened 2026-08-18
feat(attach): resolve attach endpoints per platform

EivMeyer:
Windows native support has been blocked since January by two independent problems (#70). This PR fixes the first one: attach IPC. It does **not** enable Windows — the preflight gate stays in place, because the second blocker is untouched. See *Scope* below.

## The problem

On Unix a socket path plays two roles at once:

- the address `net` binds to, and
- the on-disk record that `listAttachableTasks`, `cleanupStaleSocket`, `listAttachableClusters` and `cleanupClusterSockets` operate on via `readdir` / `existsSync` / `unlink`.

The `.sock` node is both, which is why nothing in discovery has ever needed to know about addressing.

Windows named pipes live in a `\\.\pipe\` namespace that `fs` cannot enumerate, stat or unlink. So the obvious fix — swap the path for a pipe name — silently breaks every discovery and cleanup path at once. That is the part #70's sketch doesn't cover, and it's the reason this isn't a three-line change.

## The approach

Split the two roles behind `src/attach/socket-endpoint.ts`:

| role | Unix | Windows |
| --- | --- | --- |
| record | the `.sock` socket node | a regular file at the same `.sock` path |
| endpoint | that same path | `\\.\pipe\zeroshot-<sha256(path)[0:32]>` |

Discovery, staleness cleanup and cluster teardown are untouched — they still see a file at the path they already look at. `listAttachableTasks` already accepted `entry.isFile()` alongside `entry.isSocket()`, so it matches the record as-is.

The pipe name is **derived** from the record path rather than read out of it, so every process resolves the same name with no parse step and no read/write race at startup. It's lowercased before hashing because Windows paths are case-insensitive, and the digest keeps the name inside the pipe-namespace length limit however deep the record path is.

Liveness keeps its existing shape: the record says an endpoint was claimed, a connect attempt says whether anything still serves it. A dead process leaves a record that fails to connect — exactly the stale-socket case `cleanupStaleSocket` already handles.

The seam also owns the transport. Connect sites call `connectToEndpoint()` rather than `net.createConnection()`, so an untranslated record path cannot reach the wire, and `attach-client` takes its socket type from the seam instead of importing `net` for it.

**Unix behaviour is unchanged by construction:** `endpointFor()` returns the path it was given and `writeEndpointRecord()` is a no-op.

## Scope — what this does not do

Executable resolution (#57, #457) is the other half and is **not** in this PR. `_resolveZeroshotPath()` returns a bare `'zeroshot'` string that is threaded through ~15 `spawn` / `runCommand` sites, and on Windows the npm global is `zeroshot.cmd`, which `spawn()` cannot resolve. `shell: true` is not a safe fix there because task prompts contain spaces and would be re-split into positional args. Doing it properly means a `{command, args}` refactor across `agent-task-executor.js` and `claude-task-runner.js`, which deserves its own review.

So `runPreflight()` still rejects `win32`. Nothing here changes what users can run today; it removes one of the two reasons they can't.

## Verification

- 2934 passing, 12 failing — the same 12 that fail on unmodified `main` in this environment (OMP SDK `darwin/arm64`, opcore-dependent suites). Failure set diffed byte-for-byte against a pristine `origin/main` worktree: identical.
- `tsc --noEmit` clean, `eslint` clean on changed files, opcore pre-commit gate green.
- 9 new unit tests covering both platforms: pipe-namespace shape, determinism, case-insensitivity, distinct records, long-path bounding, and that the record stays a readdir-visible, unlinkable regular file.

**Not verified on Windows.** Nobody on the team runs it, and that is the actual reason this has sat since January rather than any difficulty in the code. The unit tests pin the addressing contract on both platforms, but a real `zeroshot task run` on Windows has not happened. It cannot until #457 lands too. Reviewers should read this as a correct-by-construction seam, not as a tested port.

Refs #70, #63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -223,6 +223,7 @@ src/attach/ring-buffer.js
 src/attach/protocol.js
 src/attach/send-input.js
 src/attach/socket-discovery.js
+src/attach/socket-endpoint.js
 src/attach/attach-client.js
 src/attach/attach-server.js
 src/attach/attach-server-types.js
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -460,6 +460,7 @@ export default [
       'src/attach/protocol.js',
       'src/attach/send-input.js',
       'src/attach/socket-discovery.js',
+      'src/attach/socket-endpoint.js',
       'src/attach/attach-client.js',
       'src/attach/attach-server.js',
       'src/attach/attach-server-types.js',
--- a/src/attach/attach-client.ts
+++ b/src/attach/attach-client.ts
@@ -1,9 +1,9 @@
 /** Terminal client for attaching to a running task or agent. */
 import crypto from 'node:crypto';
 import { EventEmitter } from 'node:events';
-import net from 'node:net';
 
 import protocol from './protocol';
+import { connectToEndpoint, type EndpointSocket } from './socket-endpoint';
 
 const CTRL_B = '\x02';
 const CTRL_C = '\x03';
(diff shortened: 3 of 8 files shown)


mkceichelbeck:

PR #1058, opened 2026-08-30
fix(parser): improve extractDirectJson to handle conversational filler

VIVAAN-DHAWAN:
This PR improves the test result JSON parser (`extractDirectJson`) to handle conversational filler text (e.g. LLM outputs before/after the JSON block) without requiring markdown codeblocks. It finds the substring between the first `{` and last `}` when a direct JSON parse fails, which increases robustness.


mkceichelbeck:

